PR #185, opened 2016-07-12
Allow an author to edit node metadata, including child order

tessereth:
**DO NOT MERGE UNTIL AFTER #180**

Obviously this is not the best user experience possible but it makes it possible to change node order.

screen shot 2016-07-12 at 4 39 15 pm

### DIFF
--- a/app/controllers/editorial/nodes_controller.rb
+++ b/app/controllers/editorial/nodes_controller.rb
@@ -59,22 +59,17 @@ module Editorial
     end
 
     def edit
-
       @node = Node.find(params[:id])
       authorize! :update, @node
-      @type_name = @node.class.name.underscore
-      @form = "#{@node.class.name}Form".constantize.new(@node)
-      @editor = params[:editor]
-      redirect_to new_editorial_section_submission_path(@section, node_id: @node, editor: @editor)
+      @form = NodeMetadataForm.new(@node)
     end
 
     def update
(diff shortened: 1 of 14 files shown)


micapam:
What if one were to set all order values the same? 3, 3, 3? I'm assuming that passes validation & just results in a random order - correct?

tessereth:
Correct. Nodes with the same number will come out in database order.

micapam:
In that case, LGTM

ltankey:
LGTM on the proviso that the DB query for siblings with the same weight returns the same random order every time.

app/controllers/editorial/nodes_controller.rb
```diff
@@ -101,8 +96,7 @@ def configure_defaults!
 
     def try_save
       return false unless @form.validate(params.require(:node).permit!)
```
ltankey: Calling `@form.validate` will return a boolean expression with the result of the validation, so this method is probably unnecessary.

ltankey:
Just a final thought - should we provide the ability to add this metadata on node/page creation?

tessereth:
We currently allow setting the parent on node/page creation. But when you create a page it doesn't have any children. Possibly you could set where a page sits among it's siblings but that's a little harder (requires updating all the other siblings). It's probably a useful feature to have eventually though.

PR #186, opened 2016-07-13
Fix for sections refactor migration bug (and some other problems)

micapam:
Main fix here is that names were not being created for SectionHome nodes generated in the section refactor migration (see SITES-383).

Also a couple other fixes / cleanups:
- A revision is now created for name content attribute for each SectionHome (all modifications to content attributes require a revision to ensure history is correct)
- Order num and token set for newly-created SectionHome nodes (as these aren't using full model with callbacks, I missed this previously)
- More verbose error message for SectionHome validation (forgot to do this in SITES-168)
- SectionController class deleted (forgot to do this in SITES-168)


ltankey:
LGTM

tessereth:
LGTM

PR #187, opened 2016-07-13
Fix for editorial/nodes/show

micapam:
Recent clashing changes to templates and editorial controllers resulted in editorial/nodes/show throwing an exception (SITES-384). This fixes that.


ltankey:
LGTM

tessereth:
LGTM

PR #188, opened 2016-07-14
Fixes for various bugs from review feedback

micapam:
- root and section home node types were showing in menu
- admin/agencies/show exception
- editing section home node content (actually this is one I spotted - edit page threw an exception)

### DIFF
--- a/app/controllers/editorial/nodes_controller.rb
+++ b/app/controllers/editorial/nodes_controller.rb
@@ -27,7 +27,9 @@ module Editorial
       configure_defaults!
       authorize! :create_in, @section
 
-      @node_types = Node.descendants.map do |klass|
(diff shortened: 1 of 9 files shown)


app/controllers/editorial/nodes_controller.rb
```diff
@@ -27,7 +27,9 @@ def prepare
       configure_defaults!
       authorize! :create_in, @section
 
-      @node_types = Node.descendants.map do |klass|
+      @node_types = Node.descendants.select {|klass|
```
coderdan: reject rather than select might make this a bit clearer

coderdan:
LGTM

ltankey:
LGTM

PR #190, opened 2016-07-14
Fixed resource paths for admin sections

micapam:
As per: https://govausites.atlassian.net/browse/SITES-391
Fix: updated #find_resource method for admin department and minister controllers.

### DIFF
--- a/app/controllers/admin/departments_controller.rb
+++ b/app/controllers/admin/departments_controller.rb
@@ -17,7 +17,7 @@ module Admin
     # for more information
 
     def find_resource(param)
-      Department.find_by! slug: param
+      Department.find param
     end
   end
 end
--- a/app/controllers/admin/ministers_controller.rb
+++ b/app/controllers/admin/ministers_controller.rb
@@ -17,7 +17,7 @@ module Admin
     # for more information
 
     def find_resource(param)
-      Minister.find_by! slug: param
+      Minister.find param
     end
   end
 end
--- a/spec/controllers/admin/department_controller_spec.rb
+++ b/spec/controllers/admin/department_controller_spec.rb
@@ -6,7 +6,7 @@ RSpec.describe Admin::TopicsController, type: :controller do
 
     let(:department) { Fabricate(:department) }
 
-    subject { Admin::DepartmentsController.new.find_resource(department.slug) }
+    subject { Admin::DepartmentsController.new.find_resource(department.id) }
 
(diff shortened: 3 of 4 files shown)


app/controllers/admin/departments_controller.rb
```diff
@@ -17,7 +17,7 @@ class DepartmentsController < Admin::ApplicationController
     # for more information
 
     def find_resource(param)
-      Department.find_by! slug: param
+      Department.find param
```
tessereth: You've dropped of the `!`. Is that intensional?

micapam: ActiveRecord idiosyncracy: #find throws an exception when it can't find a record. There is no #find! method.

tobyhede: Why did this actually change from Slug?

micapam: Section#slug is deprecated, to be removed in a future migration.

micapam: (Because sections are no longer part of routing)

ltankey: It's a really good point - id vs. slug has an impact on navigation by URL, and makes things less readable. Is there a practical reason?

tobyhede: Thats what I was thinking, it makes the URLs meaningful in the admin

tessereth:
LGTM

app/controllers/admin/departments_controller.rb
ltankey: I do however think there's value in maintaining URL structure when context-switching, so that an author/reviewer can go from `/foo/bar` to `/editorial/foo/bar` and have it immediately obvious from the URL structure where they are in the hierarchy of the site. Agreed we shouldn't keep it just because we had it, but we do need to question is there value in preserving this for editorial folk.

micapam: Right, in terms of editorial, I totally agree.

micapam: Whereas this is admin.

ltankey: Yes sorry I may have hijacked this to echo the sentiment from @tobyhede as this is now happening in editorial now (not suggesting these changes did it).

maxious:
LGTM

PR #195, opened 2016-07-14
Refactored node validations into classes 

micapam:
As per SITES-375.
Also added a validator to prevent recursive ancestry (SITES-355).

### DIFF
--- a/app/models/news_article.rb
+++ b/app/models/news_article.rb
@@ -1,3 +1,6 @@
 class NewsArticle < Node
   store_attribute :data, :release_date, Date
+
+  validates_presence_of :section, :parent
+  validates :parent, ancestry_depth: { minimum: 2 }
 end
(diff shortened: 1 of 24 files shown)


app/models/news_article.rb
```diff
@@ -1,3 +1,6 @@
 class NewsArticle < Node
   store_attribute :data, :release_date, Date
-end
\ No newline at end of file
+
+  validates_presence_of :section, :parent
```
ltankey: News Articles can be created without a section, and likely won't have parents.

maxious:
LGTM

ltankey:
Aside from the validation on News Articles (which we would likely need to change in a future PR), LGMT

app/models/news_article.rb
micapam: Yeah I kind of dithered over whether to include this - and decided that I'd leave it up to the developer implementing NewsArticle to remove these validation(s) if they see fit.

micapam:
bump for status?

ltankey:
In that case LGTM

PR #197, opened 2016-07-15
Hotfix/seeds are broken

micapam:
Fixed seeds as per SITES-393.
Newly-created sections will automatically spawn a SectionHome node.

### DIFF
--- a/spec/models/node_spec.rb
+++ b/spec/models/node_spec.rb
@@ -8,7 +8,8 @@ RSpec.describe Node, type: :model do
 
   describe 'Sibling order' do
     %w(zero one two three four five).each_with_index do |num, idx|
-      let(num.to_sym) { Fabricate(:node, parent: root_node, order_num: idx) }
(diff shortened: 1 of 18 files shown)


maxious:
LGTM everything that was broken before (site home, section home, even node because of breadcrumb) is fixed after applying this patch.

A++ would merge again

spec/models/node_spec.rb
```diff
@@ -17,10 +18,11 @@
       expect(root_node.children).to eq [zero, one, two, three, four, five]
     end
 
-    let(:new_one) { Fabricate(:node, parent: root_node ) } # No explicit order num
+    let(:new_one) { Fabricate(:node, section: nil, parent: root_node ) } # No explicit order num
 
     it 'should automatically set an order number if necessary' do
       [one, two, three, new_one]
+      puts root_node.children.as_json
```
tessereth: remove debug code?

tessereth:
LGTM

PR #198, opened 2016-07-15
SITES-354 Add links between public and editorial views where appropriate

maxious:


ltankey:
This is a useful function, but should we consider putting it into the user's navbar? Viewing a published page is akin to switching contexts, which would be good to do from the top navbar.

micapam:
Yeah I agree - published view should be clean. However link to 'view published page' is fine to be on the editorial page view itself, IMO...

maxious:
I've moved it up to the top bar... which makes it obvious that you could reduce the number of links in there if you made people go to the first three links before they could access those functions (Editorial Root, This Section Editorial, This page in Editorial). Thoughts?

app/assets/stylesheets/page_controls.scss
```diff
@@ -0,0 +1,43 @@
+@media screen and (min-width: 75em) {
+[class^='controls'] {
+  section.left {
+    width: 75% !important;
```
coderdan: I'm guessing this is a temporary measure until we have the UI kit update?

app/views/application/_page_actions.html.haml
```diff
@@ -1,17 +1,26 @@
-%li
-  = link_to @section.name, editorial_section_path(@section)
-- if show_new_page_link?
-  %li
-    = link_to 'New page', prepare_editorial_section_nodes_path(@section, parent_id: @node)
+%ul.secondary-controls
+  - if node and node.section
```
coderdan: Prefer && to and

coderdan:
I know Treb is doing some more work on the visual treatment here so for now LGTM

micapam:
LGTM too.

PR #199, opened 2016-07-15
[SITES-348] Render TOC on general content pages and allow user to set TOC depth

tessereth:
screen shot 2016-07-15 at 4 18 42 pm
screen shot 2016-07-15 at 4 17 59 pm


klepas:
UI-Kit currently doesn’t provide for this — Tessa gave us a bug report (https://github.com/AusDTO/gov-au-ui-kit/issues/164).

This will likely be addressed during our next sprint.

Gemfile
```diff
@@ -24,6 +24,7 @@ gem 'jquery-rails'
 gem 'simple_form', github: 'kesha-antonov/simple_form', branch: 'rails-5-0'
 gem 'reform'
 gem 'reform-rails'
+gem 'dry-types'
```
coderdan: Are you missing strongly typed languages, @tessereth ? :p

app/forms/options_form.rb
```diff
@@ -0,0 +1,14 @@
+require 'reform/form/coercion'
+
+class OptionsForm < Reform::Form
+  feature Coercion
+  property :toc, type: Types::Form::Int,
```
coderdan: If there are only 3 options, would it be better to do something like:

    validates: { inclusion: [ 0, 1, 2] } ?

Still magic but more Ron Weasley than Harry Potter.

app/views/editorial/nodes/show.html.haml
```diff
@@ -47,6 +47,11 @@
     %tr
       %td State
       %td= node.state
+    -# TODO: refactor this to properly handle different options per Node subclass
+    - if node.options.respond_to?(:toc)
```
coderdan: Is node here decorated? If so, perhaps you could define a new method on the decorator like:

    def has_toc?
      object.options.respond_to?(:toc)
    end

app/views/editorial/nodes/options/_toc.html.haml
```diff
@@ -0,0 +1,3 @@
+-# locals: f
+= f.input :toc, collection: 0..2, label: 'Table of contents',
```
coderdan: The 0..2 feels magic. Are the numbers referring to depth?

coderdan:
A couple of minor comments but otherwise LGTM

Gemfile
tessereth: Yes. Yes I am.

app/forms/options_form.rb
tessereth: Hmmm. I remember having type related issues with using inclusion. But that may have been before I added the type cast. I'll try it again and see.

app/models/node.rb
```diff
@@ -48,6 +50,29 @@ def submission_exists_for?(user)
...
+  end
+
+  # 1) reform requires options to be an object with attributes
+  # 2) activerecord wants it to be a hash
+  # 3) Storext's default type coercion doesn't work properly with OpenStruct
+  # so for now, we just do the conversion by hand
+  def options
+    hash = super
+    if hash
+      OpenStruct.new(hash)
```
tobyhede: Is it worth actually making an options object (even just a struct)?
We have predefined options, even if they are stored in an unstructured way?

tobyhede:
LGTM, minor non-blocking feedback.

app/views/editorial/nodes/show.html.haml
tessereth: What I really want to do is iterate through all options for the given type of node. But it will be easier to figure out how to do that once we have a second option.

app/views/editorial/nodes/options/_toc.html.haml
tessereth: Yes, they do feel like magic and yes, they do refer to depth. But again, I think it will be easier to fix this up once we have a second option and the abstraction is easier to figure out.

app/models/node.rb
tessereth: Yeah, possibly. But building the proper abstractions when you only have one is hard. I'd rather wait until there's at least two.

tobyhede: Makes sense to me (MSTM)

app/models/node.rb
```diff
@@ -20,6 +20,8 @@ class Node < ApplicationRecord
   enumerize :state, in: STATES, scope: true
   content_attribute :content_body
   content_attribute :name
+  # options is not currently versioned but could be in the future
+  store_attribute :content, :options
```
micapam: Feels a bit strange to have options as part of content.... couldn't we put this in 'data'? (The other jsonb field, which NewsArticle uses to store its release date...)

micapam: If we were to version this stuff, we wouldn't be using diffs.... so might be cleaner to keep it separate even if we later extend Revision to include values for options etc...

tessereth: I put it in content under the assumption that it would be versioned eventually.

micapam: I agree, but I think the way content is versioned is different from how options will be... rather than LCS, we'd probably want to use a structural diff mechanism like https://github.com/benjamine/jsondiffpatch ...

PR #202, opened 2016-07-18
[SITES-390] Fix admin agency import bug

tessereth:
Make govcms import button its own field rather than override the collection partial which is used in multiple locations.


tobyhede:
LGTM

micapam:
LGTM too.

PR #206, opened 2016-07-18
Dealt with feedback from SITES-355.

micapam:
Validations were tested separately, but on the actual model, if you set a circular ancestry, the ancestry_depth validation would overflow. I've fixed that and introduced a spec on one of the concrete node types (GeneralContent).

See notes on https://govausites.atlassian.net/browse/SITES-355

### DIFF
--- a/app/validators/ancestry_depth_validator.rb
+++ b/app/validators/ancestry_depth_validator.rb
@@ -17,6 +17,9 @@ validates :parent, ancestry_depth: { minimum: 1, maximum: 3 }
 
 =end
 class AncestryDepthValidator < ActiveModel::EachValidator
+
+  MAX_DEPTH = 100
+
   def validate_each(record, attribute, value)
     return if record.errors.include? attribute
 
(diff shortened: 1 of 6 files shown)


maxious:
LGTM

app/validators/ancestry_depth_validator.rb
```diff
@@ -12,9 +16,10 @@
 validates :parent, ancestry_depth: { minimum: 1, maximum: 3 }
 
 =end
-
 class AncestryDepthValidator < ActiveModel::EachValidator
   def validate_each(record, attribute, value)
+    return if record.errors.include? attribute
```
tessereth: Can you be sure that the other validator will be run first? I have no idea what the validation order is.

tessereth:
Would it make sense to put a hard cap on the `get_depth` function so it can't infinite loop? If it's more than like 1000, just break out?

app/validators/ancestry_depth_validator.rb
micapam: Validators are run in the order in which they are declared in the model.

micapam:
Yeah that would be cleaner, but the problem is that ActiveRecord is trapping the problem before we even get a stack overflow.... so I suspect it would might throw an error - I can test that out. 

(Although my gut feel is this works and I want to stop thinking about it & write some actual features :)

tobyhede:
Probably don't even need 1000 max depth. Even 10 should be heaps.

tobyhede:
LGTM, nice catch

tessereth:
LGTM though if you feel like updating `get_depth` to be non-infinite it would make me even more happy.

micapam:
bump 4 status

micapam:
BTW @tessereth I added such a check for infinity.

PR #208, opened 2016-07-19
[SITES-387] Update slug when name or parent changes

tessereth:


micapam:
LGTM - though I don't understand why the slug for a dupe is `slug-` rather than slug with a whole load of alphanumeric mess after it? Was that fixed in a previous PR?

tessereth:
The slug for a dup is `slug-<uuid>`. I'm just using `#match` (regular expression match) rather than `#eq` and checking for the hyphen as a quick way of telling if it was a dup.

app/models/node.rb
```diff
@@ -73,6 +73,11 @@ def options=(value)
     super(value.to_h)
   end
 
+  def should_generate_new_friendly_id?
```
tobyhede: Where is this actually called?

tessereth: By the friendly_id magic. See http://norman.github.io/friendly_id/file.Guide.html#Deciding_When_to_Generate_New_Slugs if you want to look it up.
Actually, I could add that link as a comment if you think it will be useful.

app/models/node.rb
```diff
@@ -73,6 +73,12 @@ def options=(value)
     super(value.to_h)
   end
 
+  # See http://norman.github.io/friendly_id/file.Guide.html#Deciding_When_to_Generate_New_Slugs
```
ltankey: :+1:

ltankey:
LGTM

PR #212, opened 2016-07-20
Address feedback from QA on homepage

tobyhede:
- [x] Ministers count should use use real count from DB
- [x]  Make news real
- [x] Lose the dummy consultation

### DIFF
--- a/app/views/templates/root_node.html.haml
+++ b/app/views/templates/root_node.html.haml
@@ -53,12 +53,3 @@
           %a{href: '/ministers'}
             %span= Minister.all.count
             Ministers
-
-    .consultations
-      %h2.home-heading Public consultations
-      %ul.list-small
-        %li
-          %a
-          .meta
-            %time{datetime:'2016-05-08 00:00'} Close date 08 May 2016
-          %h3 Proposed LAP variations: Bendigo, Horsham and Shepparton radio
--- a/app/views/templates/root_node.html.haml
+++ b/app/views/templates/root_node.html.haml
@@ -29,10 +29,10 @@
 
 
   %h2.home-heading News
-  - if RootNode.news.empty?
+  - if @news.empty?
     %p No current news articles
   %ul.list-vertical--fourths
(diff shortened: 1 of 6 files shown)


app/views/templates/root_node.html.haml
```diff
@@ -29,12 +29,14 @@
 
 
   %h2.home-heading News
-
+  - if NewsArticle.published.all.empty?
```
micapam: Feels wrong to have ActiveRecord stuff in a template. Maybe have `news_articles` getter in the RootNode class instead? Same for `ministers_count`

tobyhede: There was a conversation about that on Slack. I couldn't really decide where any of this should go.

micapam: RootNode may not be optimal but better than in the presentation layer IMO...

tobyhede: @micapam Pushed the methods to RootNode.

db/seeds.rb
```diff
@@ -37,9 +37,8 @@
   section: topic,
   type: GeneralContent,
   state: :published
-}) do |node|
-  node.revise!(content_body: 'lorem ipsum').apply!
-end
+})
+node2.revise!(content_body: 'lorem ipsum').apply!
```
micapam: Why this change?

micapam:
Aside from my question on seeds.rb, this LGTM

db/seeds.rb
micapam: I think the error was the wrong node (`node` instead of `node2`) but it's a good idea to keep it in the block because if node2 already exists I guess we won't want to revise it? (Block on `find_or_create_by` only runs on create...)

tessereth:
LGTM

ltankey:
LGTM

db/seeds.rb
tobyhede: @micapam There was an error in the seeds. I think I understand what you are saying. Will address.

PR #214, opened 2016-07-20
Hard-coded infrastructure & comms category - SITES-411

micapam:
image

Rough-as-guts rip of static content from prototype.
I have set up a categories controller but it just has actions and mostly static view. 
I set up the route to match what we will likely have once we have dynamic content. 
Breadcrumb is parented off root as per prototype.

### DIFF
--- a/app/views/categories/infrastructure_and_telecommunications.html.erb
+++ b/app/views/categories/infrastructure_and_telecommunications.html.erb
@@ -1,0 +1,76 @@
+<article id="content" class="content-main">
+<% breadcrumb :category, 'Infrastructure and telecommunications' %>
+<% # n.b. All the below ripped from prototype %>
+
+<header><h1>Infrastructure and telecommunications</h1></header>
+
+<p class="abstract">Australian Government information and services related to
+  infrastructure and telecommunications.</p>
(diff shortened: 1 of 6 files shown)


app/views/categories/infrastructure_and_telecommunications.html.erb
```diff
@@ -0,0 +1,76 @@
...
+<% # n.b. All the below ripped from prototype %>
+
+<header><h1>Infrastructure and telecommunications</h1></header>
+
+<p class="abstract">Australian Government information and services related to
+  infrastructure and telecommunications.</p>
+
+<h2>Infrastructure</h2>
+
+<h4>Regional development</h4>
```
tessereth: I know you took this straight from the prototype but this should be an `<h3>` because it lives inside the `<h2>`

tessereth:
LGTM

app/views/categories/infrastructure_and_telecommunications.html.erb
micapam: Thanks @tessereth :)

ltankey:
LGTM

PR #215, opened 2016-07-20
[SITES-425] - Correctly determine if a node with a given name exists

tessereth:
Note: using find_by for name on a Node object actually queries the name column rather than `content[:name]`.

### DIFF
--- a/app/models/node.rb
+++ b/app/models/node.rb
@@ -14,6 +14,23 @@ class Node < ApplicationRecord
 
   scope :published, -> { where state: 'published' }
 
+  # These methods aren't DRY to avoid inserting
+  # unescaped strings directly into a PG query.
+  #
+  # This type of query requires the store column to be
+  # unescaped.
+  scope :with_content_value, -> (field, value) {
+    where("content ->> ? = ?", field, value)
+  }
+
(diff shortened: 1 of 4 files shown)


db/seeds.rb
```diff
@@ -16,37 +16,30 @@
 topic.summary = 'The business section covers a range of business-related topics.'
 topic.save
 
-news1 = NewsArticle.find_or_create_by!({
- name: "Business News",
+news1 = NewsArticle.with_content_value(:name, "Business News")
+            .find_or_create_by!({
```
ltankey: This indenting is a bit weird compared to the lines below... /pedant.

ltankey:
LGTM

app/models/node.rb
```diff
@@ -14,6 +14,23 @@ class Node < ApplicationRecord
 
   scope :published, -> { where state: 'published' }
 
+  # These methods aren't DRY to avoid inserting
+  # unescaped strings directly into a PG query.
+  #
+  # This type of query requires the store column to be
+  # unescaped.
+  scope :with_content_value, -> (field, value) {
```
tobyhede: Awesome. I love PG.

tobyhede:
LGTM

PR #216, opened 2016-07-20
[SITES-368] Feature/view a news article

ltankey:
This PR allows news articles that have been created via the `/editorial/news` route to be viewable under `/:section/news/:slug`. This PR also changes the construction of a news article slug to consist of `#{release_date}-#{slug}`. This prevents article clashes not just across sections, but also for identically named news articles over time.


CHANGELOG.md
```diff
@@ -7,10 +7,12 @@ This file is influenced by http://keepachangelog.com/.
 
 - Add ability to add news articles from /editorial/news
 - Provide link to add news to control bar
+- Published news articles now viewable from /:section/news/:slug
 
 ### Changed
 
 - Used ui kit via SCSS include for greater customisation
+- News articles slugs are now a combination of release dates and article names
```
tessereth: possibly also changed "News articles have no parent"

app/validators/orphaned_node_validator.rb
```diff
@@ -0,0 +1,7 @@
+class OrphanedNodeValidator < ActiveModel::Validator
+  def validate(node)
+    unless node.parent.present?
+      node.errors.add :parent, 'Nodes must have parents'
+    end
+  end
+end
```
tessereth: Could we just validate `presence: true` on `:parent` instead of having a custom validator?

app/views/editorial/news/show.html.haml
```diff
@@ -1 +1 @@
-%h1 @article.name
+%h1 #{@article.name}
```
tessereth: It's probably more idiomatic to say `%h1= @article.name`

app/views/layouts/news_article.html.haml
```diff
@@ -0,0 +1,10 @@
+= inside_layout :application do
+  = yield
+
+  = content_for :header do
+    %a.site-title{href: nodes_path(@section.home_node.path)}= @section.name
+
+  = content_for :toolbar_actions do
+    / TODO: Cleanup (perhaps in https://govausites.atlassian.net/browse/SITES-168)
```
tessereth: Remove reference to non-publicly accessible link?

app/views/templates/news_article.html.haml
```diff
@@ -1,15 +1,25 @@
...
 
   %h1
     =node.name
 
-  - if node.release_date
-    %p
-      Release Date:
-      = node.release_date
+  %p.abstract
+    =node.short_summary
```
tessereth: This may change re #218, but I'm okay with that being a separate PR.

app/views/templates/news_article.html.haml
```diff
@@ -1,15 +1,25 @@
...
+    %li
+      Published by:
+      %a= link_to(node.section.name, nodes_path(node.section.home_node.slug))
+
 
   = render partial: 'templates/content_body', object: node.content_body
 
   %p
-    =@hierarchy
+    =@hierarchy
```
tessereth: This `%p` should be removed. I think it was for debugging some time ago.

config/breadcrumbs.rb
```diff
@@ -9,7 +9,7 @@
 
 crumb :public_node do |node|
   if node.parent.present?
-    link node.name, nodes_path(path: node.path)
+    link node.name, node.full_path #nodes_path(path: node.path)
```
tessereth: remove commented out code

config/breadcrumbs.rb
```diff
@@ -35,6 +35,16 @@
...
+crumb :public_news_articles do |section|
+  link 'News', news_articles_path(section.slug)
+  parent :public_node, section.home_node
+end
+
+crumb :public_news_article do |node|
+  link node.name, news_article_path(node.section.home_node.slug, node.slug)
+  parent :public_news_articles, node.section
+end
+
```
tessereth: If I'm being pedantic, these breadcrumbs should live with the other public breadcrumbs rather than in the middle of the editorial ones.

spec/controllers/editorial/news_controller_spec.rb
```diff
@@ -45,8 +48,8 @@
         post_action
       end
 
-      it 'should link to news_home_node as parent' do
-        expect(Node.last.parent).to eq(news.home_node)
+      it 'should link to no node as parent' do
+        expect(Node.last.parent).to eq(nil)
```
tessereth: `it 'shouldn't have a parent'`?

spec/controllers/news_controller_spec.rb
```diff
@@ -0,0 +1,29 @@
...
+  let!(:user) { Fabricate(:user, author_of: section_a) }
+  let!(:article_published) { Fabricate(:news_article, section: section_a, state: 'published') }
+  let!(:article_draft) { Fabricate(:news_article, section: section_a, state: 'draft') }
+
+  describe 'GET #show' do
+    context 'when a user is not authorised' do
+      context 'for a published page' do
+        subject { get :show, section: section_a.home_node.slug, slug: article_published.slug }
+
+        it { is_expected.to be_success }
```
tessereth: Maybe also check it assigns the correct node?

spec/features/editing_content_spec.rb
```diff
@@ -56,7 +56,7 @@
     let!(:node1) { Fabricate(:general_content, state: 'published',
       parent: section1.home_node, section: section1) }
     let!(:node2) { Fabricate(:news_article, state: 'published',
-      parent: section2.home_node, section: section2) }
+      section: section2, release_date: Date.today) }
```
tessereth: `Date.today` is the default for `:release_date` and it does not appear to be explicitly referenced. Do you need to set it?

spec/validators/root_protection_validator_spec.rb
```diff
@@ -3,38 +3,38 @@
...
 
   context 'No root node exists' do
-    context 'with a parent' do
-      let(:parent) { 'something' }
+    context 'with a root node' do
+      let(:type) { 'RootNode' }
       it { is_expected.to be_valid }
     end
 
     context 'without a parent' do
```
tessereth: `without a root node`? Also `with a root node` and `without a root node` seem odd here. I think the tests are actually testing `as a root node` and `not as a root node` or something like that.

tessereth:
So I made a bunch of pedantic comments but otherwise this looks good to me. Unfortunately I helped author it so I'll let someone else be the official rubber stamp of approval.

app/validators/orphaned_node_validator.rb
ltankey: Good point.

spec/features/editing_content_spec.rb
ltankey: Didn't catch that in the fabricator, thanks!

spec/validators/root_protection_validator_spec.rb
ltankey: Yes the language is a bit confusing, as there are two contexts - the first context talking about what's already existing, and the second talking about what's being created. I might rename to make clearer.

maxious:
Could you please link the home page news articles to their full view https://github.com/AusDTO/gov-au-beta/blob/develop/app/views/templates/root_node.html.haml#L38

ltankey:
Good catch @maxious, added.

app/models/general_content.rb
```diff
@@ -3,6 +3,7 @@ class GeneralContent < Node
 
   validates_with SectionHeritageValidator
   validates_presence_of :section
+  validates :parent, presence: true
```
micapam: Isn't this redundant? As we are testing ancestry depth. Or if you want to keep it for readability purposes, you could combine the two lines as: 
`validates :parent, presence: true, ancestry_depth: { minimum: 2 }`

ltankey: Yes I suppose it might be completely redundant, happy to remove it but it does otherwise seem like it's testing two distinct things. However I can't think of an instance with >0 ancestors being tested that would fail a parent presence test. I'll remove it.

micapam:
@ltankey, if you do a search on the codebase for `Node.root` (a method from `acts_as_tree`), you'll see there are various places where it's assumes that a node with no parent is the root node. 

Since that's no longer the case with these set of changes (NewsArticles are orphans), we should add a `root` method to the Node class that just grabs the actual root node (e.g. `RootNode.first`).

tobyhede:
This is part of the reason I suggest news would be children of the news section home. We dont break the semantics of the tree.

ltankey:
@micapam that's a good clean up task and one I'll need to do. 

I think it's dubious to assume that our RootNode is reflected by the root of acts_as_tree root, as it assumes all content fits within a single node parental hierarchy. We have a more explicit way of checking (via RootNode) directly without impacting on the flexibility we have with our node trees.

app/models/news_article.rb
```diff
@@ -4,11 +4,33 @@ class NewsArticle < Node
...
+
+  def layout
+    'news_article'
+  end
+
+  # News Articles require specific redirects to a custom route that
+  # is not defined by node hierarchy. As such, this method is a convenience
+  # wrapper around the url helpers, to allow this model to override the default
+  # redirect action.
+  def full_path
```
micapam: This feels like it should be in helpers rather than in the model.

ltankey: As we use it in controllers and not just views, does it not make sense to be a model method? I thought helpers were more-or-less explicitly for view functions.

micapam: I tried writing a helper but it was ugly & used type inspection.... but I've thought of a simple way - why not just override #path_elements in NewsArticle?

````
def path_elements
  [section.home_node.slug, 'news', slug]
end
````

Then you can treat NewsArticle same as any other node.

tessereth: But then you would be calling `nodes_path(node.path)` when technically a news article isn't using that route. Using `#full_path` here also means we no longer need to remember what the helper is called or what it's parameters are. And if we need to change it in the future it's all in one place (like when we had to remove the section from every call to nodes_path).

ltankey: I'm a little lost on the distinction of the different approaches - yours would certainly work, but it doesn't really shift much more outside of the model. The model still has an implicit understanding of its path - all moving the call to the url_helper does it shift the call into the views. The actual knowledge of the path remains in the model.

Is there a Ruby idiom I'm missing that might help explain preferences to keep this kind of stuff outside of the model?

micapam: I guess it just feels wrong to be calling routing aspects of Rails inside an ActiveRecord model. 

It's not a big deal, maybe since we're trying to move fast, let's just push this through then if I can think of a clean refactor I will raise as a separate PR...

ltankey: I think I understand the distinction now: constructing data in the model (good), vs. calling url_helpers for routing (bad).

Perhaps as this is working and as you say, we're racing forward, we could have a discussion about how we would approach this as a team and raise a tech-debt card to address once we're over our hump.

maxious:
LGTM

micapam:
Caveats over routing aside, LGTM

PR #219, opened 2016-07-21
Make agencies departments again

tobyhede:


db/migrate/20160721013450_make_agencies_departments.rb
```diff
@@ -0,0 +1,10 @@
+class MakeAgenciesDepartments < ActiveRecord::Migration[5.0]
+  def up
+    sql = "UPDATE sections SET type = 'Department' WHERE type = 'Agency'"
```
tessereth: Should *ALL* Agencies be Departments? Are Agencies still a thing?

tessereth:
LGTM

ltankey:
I'm not sure of the context of this PR... Is there an associated card or something?

tobyhede:
@ltankey SITES-417
The issue is that Agencies and Depts are not the same thing. We have been using the word Agency, but the prototype has Depts with many Agencies.

micapam:
In the branch name @ltankey

ltankey:
Ahh sorry missed it in branch name.

@tobyhede @tessereth explained it to me, makes sense, LGTM.

tobyhede:
Sorry, should have carried into the PR title

micapam:
I'm just running this locally to test BTW @tobyhede

tobyhede:
Thanks @micapam Good idea. I was just running it again to double double check

micapam:
Works fine. 
I guess ideally I'd prefer this to be a rake task rather than a migration but I can't think of a scenario in which this would actually be a problem so LGTM.

PR #220, opened 2016-07-21
[wip] Display section topics on section home page

tobyhede:
Images are separate ticket. 

govaubeta

Can't see where the css is being pulled from for this layout. Might be a diff version of ui-kit


tessereth:
If you need to update ui-kit to get the right styles for this, you probably should. Just to make sure you have the right markup. I think there's a rake task for that now.

ltankey:
`rake ui_kit:update` should work.

ltankey:
Is there a card associated with this PR?

ltankey:
Changelog?

app/models/section.rb
```diff
@@ -19,6 +19,8 @@ class Section < ApplicationRecord
 
   after_create :generate_home_node
 
+  scope :alphabetically, ->  { order(:name => 'ASC') }
+
```
tessereth: we called this `:by_name` in `NewsArticle`. Should we make them the same?

tobyhede: Good idea, changed

app/views/templates/section_home.html.haml
```diff
@@ -2,3 +2,12 @@
   - breadcrumb :public_node, node
 
   = render partial: 'templates/content_body', object: node.content_body
+
+  - if @section.topics.any?
+    %h4 Popular now
```
micapam: Maybe this should say 'Information and services' ? I think that's the important bit rather than 'popular now' (popularity isn't implemented yet anyway).

tobyhede:
@micapam I've just copied the prototype. The information and services bit has content, so that probably needs to be part of the editable content for the page.

app/views/templates/section_home.html.haml
```diff
@@ -2,3 +2,12 @@
...
+    %ul.list-vertical--thirds
+    - @section.topics.by_name.each do |topic|
```
tessereth: Do only agencies have topics? If so does this template work for all section home pages? Also I tried to test it and there's no way in admin to add a topic to an agency. It might be worth doing that as well.

tobyhede:
This is now wip pending some alignment on what we're actually building

PR #222, opened 2016-07-21
[SITES-424] Add a node type for pages with custom templates

tessereth:
This is required scaffolding for the times and dates content, which will be added in future PRs.

### DIFF
--- a/app/controllers/editorial/nodes_controller.rb
+++ b/app/controllers/editorial/nodes_controller.rb
@@ -27,9 +27,7 @@ module Editorial
       configure_defaults!
       authorize! :create_in, @section
 
-      @node_types = Node.descendants.reject {|klass|
-        [RootNode, SectionHome].include? klass
-      }.map do |klass|
+      @node_types = [GeneralContent].map do |klass|
         name = klass.name.underscore
         [I18n.t("domain_model.nodes.#{name}"), name]
       end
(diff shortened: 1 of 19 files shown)


app/controllers/editorial/nodes_controller.rb
```diff
@@ -27,9 +27,7 @@ def prepare
       configure_defaults!
       authorize! :create_in, @section
 
-      @node_types = Node.descendants.reject {|klass|
-        [RootNode, SectionHome].include? klass
-      }.map do |klass|
+      @node_types = [GeneralContent].map do |klass|
```
maxious: Do we still need to exclude RootNode, SectionHome etc. further in the processing

micapam: Do we even need to have the node type selector if there's only one choice....?

tessereth: I was running under the assumption that we would add more types at some stage so I didn't want to remove the whole prepare page/selector thing.

ltankey: I think there's only one choice right now. When we grow the number of node types we need to provide the ability to create, this will grow into presumably a more complex solution.

micapam:
@tessereth, what's the sequence of events to get one of those time and date pages into the system? I'm not quite clear on how this works.

tessereth:
* Add the times and dates pages as html templates (in `views/templates/custom`)
* In admin, create a CustomTemplateNode node with the correct template

maxious:
LGTM (but This branch has conflicts that must be resolved)

micapam:
I have a bad feeling about this. But I can't think of any alternatives, so LGTM

PR #223, opened 2016-07-21
Static categories

micapam:
Static categories on the homepage as per the prototype. 
I've put the category info into a YAML file which populates an array of OpenStruct objects - so we should later be able to swap out the static data for model instances (once we implement categories in the DB).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -64,6 +64,9 @@ This file is influenced by http://keepachangelog.com/.
 - Support for PaaS instance of govCMS
 - Ability to edit page order
 - Give user option to update page parent
+- Ministers index page
+- Infrastructure and Telecommunications category (static)
+- Categories (static) shown on the homepage
 
 ### Changed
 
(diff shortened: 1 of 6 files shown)


CHANGELOG.md
```diff
@@ -64,6 +64,9 @@ This file is influenced by http://keepachangelog.com/.
 - Support for PaaS instance of govCMS
 - Ability to edit page order
 - Give user option to update page parent
+- Ministers index page
+- Infrastructure and Telecommunications category (static)
+- Categories (static) shown on the homepage
```
tessereth: These should be added at the top of the changelog, in the unreleased section.

tessereth:
Once changelog is fixed, this LGTM.

CHANGELOG.md
ltankey: I actually think they might already be there, having been added in a rebase (except for the last one).

micapam:
Ah I was reading the changelog upside-down :)

ltankey:
LGTM

PR #224, opened 2016-07-21
fix viewing news tags on article pages

ltankey:
Fixes the view of the distribution tags on news articles to look right:

image

Also improved the semantic markup for the `/news` page's usage of definition lists.

### DIFF
--- a/app/views/news/index.html.haml
+++ b/app/views/news/index.html.haml
@@ -17,5 +17,5 @@ Announcements, media releases, interviews, speeches and more from the Australian
             %dl
               %dt
                 Topics:
-                - article.sections.each do |section|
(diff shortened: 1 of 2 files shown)


micapam:
Is it necessary to repeat the tag for the currently-viewed section?

ltankey:
Oh, right - in that instance the publisher is included in the tag.

Well, probably not - but I think that's because I selected it from the distribution options when I created the NewsArticle. You're right in that this is probably not something we want - but the question is should it belong in the presentation layer or in the model validation/sanitsation?

tessereth:
LGTM

tobyhede:
LGTM

ltankey:
rebased

PR #227, opened 2016-07-21
remove redundant class on <main>

simonschwartz:
There was a class on the `<main>` container that was removing the top padding from the UI Kit.

**Before**

bad

**After**

good


maxious:
Oh. LGTM

micapam:
Ah I was wondering about that! LGTM.

maxious:
LGTM

PR #228, opened 2016-07-21
[SITES-424] Add select box for setting the custom template node's template

tessereth:
Also remove unnecessary custom administrate fields that can actually be done using `Field::Select`.


micapam:
LGTM !

coderdan:
LGTM

maxious:
LGTM

PR #231, opened 2016-07-22
remove published scope which was messing with RootNode resolution

ltankey:
Removes the published scope to make RootNode.first work as intended.

### DIFF
--- a/app/controllers/nodes_controller.rb
+++ b/app/controllers/nodes_controller.rb
@@ -6,7 +6,7 @@ class NodesController < ApplicationController
 
   def show
 
-    @node = Node.published.find_by_path! params[:path] || ''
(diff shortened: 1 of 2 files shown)


app/controllers/nodes_controller.rb
```diff
@@ -6,7 +6,7 @@ class NodesController < ApplicationController
 
   def show
 
-    @node = Node.published.find_by_path! params[:path] || ''
+    @node = Node.find_by_path! params[:path] || ''
```
coderdan: Won't this allow the viewing of any node publicly regardless of whether or not it has been published?

coderdan:
LGTM

micapam:
LGTM

PR #238, opened 2016-07-22
Add section image_url column and admin field

maxious:
This will add a column and admin fields to edit an Image URL for a topic/section/minister/department. But won't provide editorial view or any public facing view yet until their are places to use this field.


ltankey:
LGTM

micapam:
I get that you don't want to allow these to be edited in editorial... but couldn't we at least edit them in admin ?

micapam:
Ah never mind, I see you've done that already :) LGTM

PR #241, opened 2016-07-25
Display news on Section home pages

tobyhede:
If anyone has a good idea on how to test something like this ... would a feature be best?

govaubeta

### DIFF
--- a/app/controllers/editorial/nodes_controller.rb
+++ b/app/controllers/editorial/nodes_controller.rb
@@ -18,8 +18,10 @@ module Editorial
 
     def show
       @node = Node.find(params[:id])
-      @section = @node.section      
-      @news = NewsArticle.published_for_section(@section).limit(3)
+      @section = @node.section
+      if @section
+        @news = NewsArticle.published_for_section(@section).limit(3)
+      end
       set_menu_nodes
     end
 
--- a/app/controllers/editorial/nodes_controller.rb
+++ b/app/controllers/editorial/nodes_controller.rb
@@ -18,6 +18,8 @@ module Editorial
 
     def show
       @node = Node.find(params[:id])
+      @section = @node.section
+      @news = NewsArticle.by_section(@section).by_release_date.by_published_at.published.limit(3)
       set_menu_nodes
     end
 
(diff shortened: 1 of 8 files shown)


app/controllers/editorial/nodes_controller.rb
```diff
@@ -18,6 +18,8 @@ def index
 
     def show
       @node = Node.find(params[:id])
+      @section = @node.section
+      @news = NewsArticle.by_section(@section).by_release_date.by_published_at.published.limit(3)
```
tessereth: For testing, maybe you could make this a method/scope on NewsArticle and unit test that?

tobyhede: Extracting a method is good idea.

spec/features/section_home_page.rb
```diff
@@ -0,0 +1,21 @@
+require 'rails_helper'
+
+RSpec.describe 'home page', type: :feature do
+
+  Warden.test_mode!
+
+  let!(:root_node) { Fabricate(:root_node) }
```
tessereth: I think I've fixed the fabricators so you don't need to create the root node explicitly anymore.

tessereth:
changelog?

app/models/news_article.rb
```diff
@@ -20,8 +20,16 @@ class NewsArticle < Node
...
   }
 
+  scope :by_section, -> (section) {
+    where(:section_id => section.id)
+  }
+
   validates_presence_of :section
 
+  def self.published_for_section(section)
+    NewsArticle.by_section(section).by_release_date.by_published_at.published
```
tessereth: `by_release_date` and `by_published_at`. It seems odd to sort by both. The release date is the one the user can choose so that's probably the more relevant thing to sort by.

tessereth:
LGTM

ltankey:
Do we need section news home to display items for which the section is _not_ the publisher, but is only referenced via the distribution list?

tobyhede:
@ltankey I have assumed no. The prototype shows direct news, speeches etc.

ltankey:
Awesome then, LGTM

ltankey:
I know this is now closed, but a thought occurred: was there a reason behind not using `@section.news_articles` instead of a method on NewsArticle? I know we discussed this earlier, but that was in the context of distribution lists too.

tobyhede:
@ltankey section.news_articles doesnt return the news published by the section.

ltankey:
Right, sorry, that's a bummer. I missed that point.

PR #243, opened 2016-07-25
Remove unnecessary administrate view override

tessereth:
Remove file that overrides normal administrate behaviour but is exactly the same as the original. Basically I forgot to remove it in #202.


maxious:
LGTM

ltankey:
LGTM

PR #245, opened 2016-07-25
SITES-418 `public_node_path` helper

freshtonic:
- removed `Node#full_path`
- subclasses override `Node#path_elements` if they need custom routing
- all nodes paths generated using a helper that hides *how* path is generated

  e.g. `public_node_path(news_article)`

`public_node_path` is the publically-routeable path for the node. It's
currently defined as `nodes_path(path: node.path)`.

### DIFF
--- a/app/controllers/editorial/news_controller.rb
+++ b/app/controllers/editorial/news_controller.rb
@@ -1,5 +1,6 @@
 module Editorial
   class NewsController < EditorialController
+    include ::NodesHelper
     before_action :set_sections_by_membership, only: [:index, :new, :create, :edit]
 
     def index
@@ -79,7 +80,7 @@ module Editorial
             @node.save!
           end
 
-          redirect_to @node.full_path
+          redirect_to public_node_path(@node)
           return
         end
       end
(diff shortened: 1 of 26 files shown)


ltankey:
This looks great. Do we want to refactor out usage of `node_path` too? Or do these changes cover off all usages of that URL helper?

micapam:
I like the implementation - better than previously in fact because we can just pass node to `public_node_path` rather than node.path. 

However if you search the codebase for `nodes_path` you find a few instances that could be replaced ... 

That's a quibble though, LGTM.

ltankey:
LGTM pending refactor of `node_path`

freshtonic:
Quibbles addressed. 😃

app/models/section_home.rb
```diff
@@ -3,4 +3,7 @@ class SectionHome < Node
   validates :parent, ancestry_depth: { equals: 1 }
   validates :section, presence: true, uniqueness: true
 
+  def path_elements
+    [section.home_node.slug]
+  end
```
tessereth: Do section home nodes actually need an override here? I thought they were working fine without it.

freshtonic: oh wait, that makes no sense at all.

tessereth:
LGTM

PR #246, opened 2016-07-25
[SITES-432][SITES-439] Improved navigation for times and dates custom templates

tessereth:
The times and dates pages need to suppress the state-specific pages appearing in the primary nav and the breadcrumbs. The easiest way to do that is to add an option on the node that is currently only settable through admin.

image


ltankey:
LGTM

micapam:
LGTM

maxious:
LGTM

PR #247, opened 2016-07-25
[SITES-287] Feature/summaries at top of pages

ltankey:
image

Add short_summary to meta description for SEO also.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -40,6 +40,7 @@ This file is influenced by http://keepachangelog.com/.
 - Changed Table of Contents heading to 'On this page'
 - News Articles get sorted by published_at date
 - News Articles short summary field can be edited
+- Short/long summary fields for all nodes sit beneath page abstract
 
 ### Fixed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -40,7 +40,7 @@ This file is influenced by http://keepachangelog.com/.
 - Changed Table of Contents heading to 'On this page'
 - News Articles get sorted by published_at date
 - News Articles short summary field can be edited
-- Short/long summary fields for all nodes sit beneath page abstract
+- Short/long summary fields for all nodes sit within page abstract
 
 ### Fixed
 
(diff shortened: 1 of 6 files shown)


CHANGELOG.md
```diff
@@ -32,6 +32,7 @@ This file is influenced by http://keepachangelog.com/.
 - Changed Table of Contents heading to 'On this page'
 - News Articles get sorted by published_at date
 - News Articles short summary field can be edited
+- Short/long summary fields for all nodes sit beneath page abstract
```
tessereth: beneath page abstract? I thought they were the page abstract?

ltankey: Language fail - beneath for HAML - within from a markup perspective. Will modify for clarity.

app/views/templates/general_content.html.haml
```diff
@@ -1,8 +1,17 @@
+- if node.short_summary
```
micapam: Is this not the case across the board? Could it go in application layout?

ltankey: It is, but some nodes may not have a defined short summary (i.e. SectionHome nodes).

With that in mind, would it still make sense for us to shift this into app layout?

app/views/application/_head.html.erb
```diff
@@ -1,6 +1,7 @@
 <head>
   <title><%= title %></title>
   <meta name="viewport" content="width=device-width, initial-scale=1">
+  <%= yield :meta_description %>
```
tessereth: Does this need an `if content_for? :meta_description`? I actually have no idea what happens if you yield to something that isn't defined.

app/views/templates/general_content.html.haml
micapam: You are already checking to see if there is a short summary though aren't you? The conditional above.

ltankey: Good point - I'll move it into the application layout.

ltankey:
@micapam I shifted the content_for block into the node layouts, which felt like a better place than their templates. I couldn't move them into the application layout as that makes an assumption that every page is represented by a node, which isn't necessarily true.

micapam:
LGTM!

tessereth:
LGTM

PR #248, opened 2016-07-26
[SITES-445] Correctly determine who sees the review buttons

tessereth:


app/helpers/submissions_helper.rb
```diff
@@ -1,6 +1,6 @@
 module SubmissionsHelper
   def can_review?
```
freshtonic: Is this helper necessary? I _think_ you could use the callback style of writing the ability and do the `submission.section.present?` logic inside the block.

tessereth: That's a fair point. Will move the logic into `ability.rb`.

tessereth: done

maxious:
LGTM

ltankey:
LGTM

PR #249, opened 2016-07-26
Use simpler singleton pattern in Metrics recorder

maxious:
Tiny change but this should improve stability


tessereth:
LGTM

freshtonic:
@maxious what was the actual issue that this fixes?

maxious:
I haven't seen it in Rollbar but in development Luke and I were getting "Prometheus::Client::Registry::AlreadyRegisteredError - revisions_submitted has already been registered"

ltankey:
Yes it seemed like a bizarre issue that occurred from an update to the codebase without bouncing the server. Many issues seem to happen in such a scenario, but it's good if this PR solves this particular issue.

LGTM

freshtonic:
@ltankey @maxious it sounds like a race condition across multiple threads perhaps.

app/services/metrics_recorder.rb
```diff
@@ -1,8 +1,9 @@
 require 'prometheus/client'
-require 'singleton'
 
 class MetricsRecorder
-  include Singleton
+  def self.instance
```
freshtonic: This isn't thread-safe.

freshtonic: Instead of lazy-instantiation, use a class variable that is immediately initialized and return that in `instance`.

freshtonic:
If this is failing mostly in dev there may be some interplay between autloading.

PR #250, opened 2016-07-26
Links from dept homepages to related ministers

micapam:
Links from department homepages to related ministers. Administrable via admin.

Still to do is to include the name of the minister next to the link (as per prototype).


app/views/templates/_related_ministers.html.haml
```diff
@@ -0,0 +1,13 @@
+- if related_ministers.present?
```
ltankey: Does passing in the local variable `object` automatically set a variable to the name of the template?

micapam: Yes.

tessereth: yes

ltankey:
LGTM

app/views/templates/_related_ministers.html.haml
```diff
@@ -0,0 +1,13 @@
...
+  %h2 Our ministers
+  %ul.list-horizontal
+    - related_ministers.each do |minister|
+      %li
+        %figure
+          %img{src: minister.image_url}
+        %article
+          %h3
+            =link_to(minister.name, nodes_path(path: minister.home_node.path))
```
ltankey: Should this be `public_node_path`?

micapam: Indeed!

tessereth:
LGTM

PR #253, opened 2016-07-26
fix rendering of abstract on base node template

ltankey:
Fix the metatag wrapping for the description based on short_summary of a node.

### DIFF
--- a/app/views/application/_head.html.erb
+++ b/app/views/application/_head.html.erb
@@ -1,7 +1,7 @@
 <head>
   <title><%= title %></title>
   <meta name="viewport" content="width=device-width, initial-scale=1">
-  <%= yield :meta_description %>
+  <meta name="description" content="<%= yield :meta_description %>">
   <script>
     (function (h) {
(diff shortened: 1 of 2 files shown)


micapam:
The changes LGTM. Though I'm not clear on your description - not sure what you mean by Minister? Isn't Minister a Section rather than a Node...?

ltankey:
@micapam I'm not clear on this now - I thought I tested earlier and this fixed it, but now I've realised it's redundant.

The necessary templates *should* already have the required summaries. I'll investigate further.

ltankey:
OK the changes to the template are in develop and will be merged in next release.

This PR now officially just fixes the metatag wrapping.

tessereth:
changelog?

ltankey:
The metadata bug was introduced in this change (it currently renders out the short_summary, yuck!). Do you think we need to record the fact we now have a metadata tag? Happy to put it in.

tessereth:
right. LGTM.

PR #255, opened 2016-07-26
[SITES-419][SITES-427] delegate section name and slug to its home node

tessereth:
There's a bunch of bootstrapping issues with trying to delegate the name between the section and its home node. So this does it halfway. It's not great but it works.

### DIFF
--- a/app/models/section.rb
+++ b/app/models/section.rb
@@ -13,8 +13,6 @@ class Section < ApplicationRecord
         foreign_key: :section_id,
         association_foreign_key: :connection_id
 
-  validates :name, uniqueness: { case_sensitive: false }
-
   # Note: resourcify must be called in every subclass so rolify will work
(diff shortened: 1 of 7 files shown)


app/models/section.rb
```diff
@@ -13,7 +13,7 @@ class Section < ApplicationRecord
         foreign_key: :section_id,
         association_foreign_key: :connection_id
 
-  validates :name, uniqueness: { case_sensitive: false }
```
ltankey: Do we still have anything that checks for the uniqueness of the `:name` field?

tessereth: Nope. We can validate name uniqueness on the sectionhome node but I wasn't sure if we wanted that or not.

maxious:
LGTM

ltankey:
LGTM

PR #257, opened 2016-07-26
Update changelog post 0.6.0 release

tessereth:


ltankey:
I'm not sure how we got these confused :/ LGTM though

tessereth:
The ones that were under 0.6.0 were in PR when we made the release. So after the release, we merged and they got added to the wrong section of the changelog.

Issue #259, opened 2016-07-26
Summary field labels


michaelhugill:
Hi @ltankey, looks like something we discussed on Slack didn't quite make it to Github and the platform.

I've noticed that the field labels in the CMS are 'short summary' and 'summary', but we were hoping for 'short summary' and 'long summary'. 

The back story here is that we don't want any confusion for authors so we thought it important that both fields have distinct names, ie short and long. 

Would it be easy to make that change?

ltankey:
Hi @michaelhugill thanks for raising this - it will be addressed in the next release.

michaelhugill:
Not critical (but worth recording): @ltankey I've noticed that the 'long summary' field is no longer present in CMS pages. Is this something that will come back later?

'Short summary' is there and working fine.

ltankey:
What page is this happening for @michaelhugill? I've just checked where it was added and it should still be working. Does the page have content for long summary?

ltankey:
Long summary has dropped off due to a bug.

ltankey:
Delivered in v0.9.0

PR #260, opened 2016-07-26
Fix/summary labels and sectionhome summary

### LINKED ISSUE #258: Short summary and sectionhomes aka landing pages
I’ve noticed the sectionhomes aka landing pages in the CMS don’t have the ‘short summary’ field which means that authors would be forced to place their short summaries inside the body text area (and potentially miss the styling associated with the short summary field).

At this stage I can't see an argument for a sectionhome aka landing page having a summary (aka long summary) field, but so far all minister and department landing pages definitely need short summaries visible to users.

ltankey:
Fixes #258 and #259.

### DIFF
--- a/app/views/editorial/submissions/_section_home.html.haml
+++ b/app/views/editorial/submissions/_section_home.html.haml
@@ -1,0 +1,1 @@
+= f.input :short_summary
--- a/app/views/editorial/submissions/_section_home.html.haml
+++ b/app/views/editorial/submissions/_section_home.html.haml
@@ -1,1 +1,0 @@
-= f.input :short_summary
--- a/spec/controllers/news_controller_spec.rb
+++ b/spec/controllers/news_controller_spec.rb
@@ -52,9 +52,9 @@ RSpec.describe NewsController, type: :controller do
         expect(assigns(:articles)).to_not include(article_unpub)
       end
 
-      it 'should return a list ordered by date and name' do
-        expect(assigns(:articles).to_a).to eq([article_today_c, article_today_a, article_today_b, article_yesterday])
-      end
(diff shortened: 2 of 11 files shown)


spec/controllers/news_controller_spec.rb
```diff
@@ -52,9 +54,9 @@
         expect(assigns(:articles)).to_not include(article_unpub)
       end
 
-      it 'should return a list ordered by date and name' do
-        expect(assigns(:articles).to_a).to eq([article_today_c, article_today_a, article_today_b, article_yesterday])
-      end
+     it 'should return a list ordered by date and name' do
+       expect(assigns(:articles).to_a).to eq([article_today_c, article_today_a, article_today_b, article_yesterday])
+     end
```
tessereth: I think the indenting here was originally correct and is now one space too short.

ltankey:
Fixed!

spec/controllers/news_controller_spec.rb
ltankey: Fixed!

app/views/editorial/submissions/_section_home.html.haml
```diff
@@ -0,0 +1 @@
+= f.input :short_summary
```
tessereth: Is short_summary an attribute of Node? If so, should the form entry be in `new.html.haml` rather than the subclass specific partials?

ltankey: Good point, I've moved it into the new.html.haml file.

maxious:
LGTM

tessereth:
LGTM

PR #269, opened 2016-07-27
Submissions usability fixes

micapam:
SITES-253 Show all submissions on submissions index
SITES-462 Show email if name is missing for user

Some quick fixes to make submissions more useful now that content editors want to allow only certain users to review submissions rather than everyone self-approving. 

The collection shown by editorial/submissions/index wasn't very useful (only showed user's own submissions) - this PR changes it to show all submissions in the section. 

Also, `UserDecorator#full_name` wasn't very useful for the users on our system as none of them have names. (There's no name fields in the signup form, nor in the admin UI). To allow content reviewers to see who has authored submissions, I've changed this to `display_name` and (for the time being) it shows the email if there's no name available.

### DIFF
--- a/app/controllers/editorial/submissions_controller.rb
+++ b/app/controllers/editorial/submissions_controller.rb
@@ -10,7 +10,7 @@ module Editorial
 
 
     def index
-      @submissions = scope.open.for(current_user)
+      @submissions = scope.open
     end
 
     def create
--- a/app/controllers/editorial/submissions_controller.rb
+++ b/app/controllers/editorial/submissions_controller.rb
@@ -10,7 +10,12 @@ module Editorial
 
 
     def index
-      @submissions = scope.open
+      @submissions = scope.open.recent
+
+      # Filter submissions to current user's unless s/he is a reviewer
+      unless can? :review_in, @section
+        @submissions = @submissions.for(current_user)
+      end
     end
 
     def create
--- a/app/decorators/request_decorator.rb
+++ b/app/decorators/request_decorator.rb
@@ -2,7 +2,7 @@ class RequestDecorator < Draper::Decorator
   decorates_association :approver, with: UserDecorator
   delegate_all
   delegate :name, to: :section, prefix: true
-  delegate :full_name, to: :approver, allow_nil: true, prefix: true
+  delegate :display_name, to: :approver, allow_nil: true, prefix: true
 
   def open?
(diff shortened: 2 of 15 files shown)


tessereth:
LGTM

app/controllers/editorial/submissions_controller.rb
```diff
@@ -10,7 +10,7 @@ class Editorial::SubmissionsController < Editorial::SectionsController
 
 
     def index
-      @submissions = scope.open.for(current_user)
+      @submissions = scope.open
```
ltankey: Would this potentially make the submissions index grow quite largely and useless? Particularly if not sorted in any meaningful way?

maxious:
LGTM

app/controllers/editorial/submissions_controller.rb
micapam: We didn't have sorting previously though did we? Or was that sort hidden inside `for` scope / method....? (looking it up)

ltankey: No you're right - I think it was only restricted by user. If we're to expose all submissions though, there isn't much a user can do with any but their own. Perhaps we have a lack of clarity around the intent of this page.

ltankey:
LGTM minor comments aside.

app/controllers/editorial/submissions_controller.rb
micapam: I think you mean 'there isn't a much an **author** can do with any but their own' which is a fair point. 

I can make the controller a bit smarter so it returns only actionable submissions in the context of the currently logged in user's role for the current section. 

BTW, `for` scope did indeed have a hidden order inside it... I'm going to shift that out into its own scope.

app/decorators/request_decorator.rb
```diff
@@ -14,7 +14,7 @@ def closed?
 
   def actioned_status
     if closed?
-      "This request was #{object.state} at #{actioned_at} by #{approver_full_name}"
+      "This request was #{object.state} at #{actioned_at} by #{approver_display_name}"
```
bjgordon: Maybe also localise actioned_at?

ltankey:
LGTM

app/decorators/request_decorator.rb
micapam: Good idea but out of scope for this ticket.

PR #270, opened 2016-07-27
Remove slug from section dashboard

tobyhede:
Sections no longer have slugs, remove from admin to avoid confusion


tessereth:
LGTM

PR #272, opened 2016-07-27
[SITES-403 | SITES-467] Feature/section news index

ltankey:
This PR moves us closer to phasing out news controller & routes.

The details of next steps are described in SITES-466.

This PR will use the `/news` and `/:section/news` routes to display news listings, with the latter filtering by section. For this to be displayed on a page's navbar, an author must create a general content node with the slug of 'news' - in most instances _this has already been done_, which is why this solution assumes this.

Moving forward, these GeneralContent nodes will be phased out for a specific NewsAggegator type. In the interim, the already-existing nodes provide a nice way of exposing the route for a given section, which `routes.rb` takes over.

This PR also includes some foundational plumbing in `section.rb` to manage the create action of a news aggregator on section creation.

### DIFF
--- a/app/controllers/editorial/news_controller.rb
+++ b/app/controllers/editorial/news_controller.rb
@@ -33,7 +33,7 @@ module Editorial
         if current_user.is_member?(Section.find(@form.section_id))
           # TODO: move this into a form validator
           # Prevents distribution list being created for publisher
-          @form.section_ids.reject! do |s_id|
+          @form.section_ids.reject!(&:blank?).reject! do |s_id|
             s_id == @form.section_id ||
                 !current_user.is_member?(Section.find(s))
           end
@@ -68,7 +68,7 @@ module Editorial
 
       if @form.validate(params.require(:node))
         if current_user.is_member?(Section.find(@form.section_id))
-          @form.section_ids.reject! do |s|
+          @form.section_ids.reject!(&:blank?).reject! do |s|
             s == @form.section_id.to_s ||
                 !current_user.is_member?(Section.find(s))
           end
--- a/app/controllers/editorial/news_controller.rb
+++ b/app/controllers/editorial/news_controller.rb
@@ -33,8 +33,8 @@ module Editorial
         if current_user.is_member?(Section.find(@form.section_id))
           # TODO: move this into a form validator
           # Prevents distribution list being created for publisher
-          @form.section_ids = @form.section_ids.reject(&:blank?).reject! do |s_id|
-            s_id == @form.section_id ||
+          @form.section_ids.reject! do |s_id|
+            s.blank? || s_id == @form.section_id ||
                 !current_user.is_member?(Section.find(s))
           end
 
@@ -68,9 +68,10 @@ module Editorial
 
       if @form.validate(params.require(:node))
         if current_user.is_member?(Section.find(@form.section_id))
-          @form.section_ids = @form.section_ids.reject(&:blank?).reject! do |s|
-            s == @form.section_id.to_s ||
-                !current_user.is_member?(Section.find(s))
+          @form.section_ids.reject! do |s|
+            s.blank? ||
+              s == @form.section_id.to_s ||
+              !current_user.is_member?(Section.find(s))
           end
 
           @form.save do |params|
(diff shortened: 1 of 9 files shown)


app/controllers/editorial/news_controller.rb
```diff
@@ -33,7 +33,7 @@ def create
         if current_user.is_member?(Section.find(@form.section_id))
           # TODO: move this into a form validator
           # Prevents distribution list being created for publisher
-          @form.section_ids.reject! do |s_id|
+          @form.section_ids = @form.section_ids.reject(&:blank?).reject! do |s_id|
             s_id == @form.section_id ||
                 !current_user.is_member?(Section.find(s))
```
tessereth: It seems odd to have two reject blocks chained together here.

app/controllers/news_controller.rb
```diff
@@ -6,6 +6,11 @@ class NewsController < ApplicationController
 
   def index
     @articles = NewsArticle.published.by_release_date.by_published_at
+
+    if params[:section].present?
+      set_section
+      @articles = @articles.by_section(@section)
```
tessereth: Does `:by_section` sort by date etc?
Should the other assign to @articles be in an else block so the news isn't loaded from the database twice?

ltankey: I believe scopes are lazily applied, and the `by_section` scope is just added on to the already-existing results. I could be wrong on this.

app/models/section.rb
```diff
@@ -44,6 +44,10 @@ def name
     end
   end
 
+  def news_node
+    nodes.with_name('News')
```
tessereth: This will return a list of nodes rather than a single one, which is not what the name implies

config/breadcrumbs.rb
```diff
@@ -12,6 +12,10 @@
   parent :public_node, Node.root_node
 end
 
+crumb :public_root do
+  link 'Home', root_path
+end
+
```
tessereth: We don't have a `:public_root` crumb anymore. We have `:public_node, Node.root_node` instead.

ltankey: `:public_root` is defined at line 15.

app/controllers/news_controller.rb
tessereth: Ah. I didn't read it correctly. That makes sense.

config/breadcrumbs.rb
```diff
@@ -94,12 +94,22 @@
...
 end
 
 crumb :public_news_article do |node|
   link node.name, public_node_path(node)
-  parent :public_news_articles, node.section
+  parent :public_news_articles
+end
+
+crumb :section_news_articles do |section|
+  link 'News', nil
```
tessereth: On the off chance that eventually there are children of this crumb, we should specify the path for the link here. Similarly for public_news_aggregator.

Also, what's the difference between public_news_articles and public_news_aggregator? Are they actually the same?

app/models/section.rb
```diff
@@ -61,6 +67,13 @@ def generate_home_node
       end
     end
   end
+
+
+  def generate_news_node
+    unless news_node.present?
+      # TODO: Create NewsNode
```
maxious: When does this get called? Or what would call it?

ltankey: This would be setup as an after_create hook, similar to how generate_home_node is currently called.

tessereth:
LGTM

app/models/section.rb
micapam: Maybe you could combine both into `generate_indispensible_nodes` or something

ltankey: Yeah a more generic version might be good. I'll create the generic indispensable version and leave refactoring generate_home_nodes as a future task.

micapam:
I'm not sure why the NewsAggregator stuff isn't included in this PR though @ltankey? Is it because of the existing news nodes?

micapam:
Or is it a blue/green thing?

ltankey:
To a degree yes, there's a data task at hand for moving/creating/validating the types @micapam.

Importantly, landing the functionality it isn't needed all at once. SITES-466 paints more detail around this, but there's a host of functionality we'll want around NewsAggegator that, in the context of section news index, makes less sense to bundle in this PR.

As we already have a controller and route, I'm trying not to overload the intent of this card. This will move us closer to phasing it out however.

micapam:
Good points well made, in view of them, this LGTM

PR #274, opened 2016-07-27
[SITES-332 | SITES-333] Allow setting of user names when signing up and through admin

tessereth:
This also adds the current username to the top right because otherwise I can't test that the username is actually being set. And it seems like a useful thing to display.

image


tessereth:
So technically this is in our backlog as a story and we're in feature freeze but it was easy to implement and was on the borderline of being a bug. And it seems unlikely to break things.

ltankey:
I know there were lots of conversations on the topic when we first looked at this card that looked at having simply a single name field. As we're using `display_name` in most cases now to use the email where no name is set, would it be worth us changing these name fields now? Or is there a hidden amount of work tangled in with devise that would make this non-trivial?

tessereth:
The main reason I didn't was because feature freeze. The db migration is a little fiddly and is far more likely to go wrong than just adding fields to admin and the sign up form. We should change to just 'name' eventually though.

ltankey:
LGTM

micapam:
LGTM 2

PR #277, opened 2016-07-27
SITES-438 show all submitted fields in submission

freshtonic:
Name & Body are shown first and then all remaining fields are iterated over in lexicographical order.

If we want to control the order of the fields then we'll need to create a union of all the fields of all node types and manually specify the order in the template. And we'd need to remember to update the template whenever a new field is added.

### DIFF
--- a/app/models/revision_content.rb
+++ b/app/models/revision_content.rb
@@ -19,6 +19,18 @@ class RevisionContent
     end
   end
 
+  def main_keys_in_order
+    [:name, :short_summary, :summary, :content_body]
+  end
+
+  def optional_keys_in_order
+    node.all_content.keys.reject do |key|
+      get_content(key).nil?
+    end.reject do |key|
+      main_keys_in_order.include?(key)
+    end.sort
+  end
+
   def get_content(content_key)
     return nil unless traversal_sequence.any? { |rev| rev.diffs[content_key].present? }
 
@@ -44,6 +56,10 @@ class RevisionContent
     else
       super
     end
+  end
+
+  def [](key)
+    get_content(key)
   end
 
   private
--- a/app/views/editorial/submissions/_content.haml
+++ b/app/views/editorial/submissions/_content.haml
@@ -6,14 +6,38 @@
   %dl
     %dt Name
     %dd
-      - if content.name.nil?
+      - if content[:name].nil?
         = defaults[:name]
       - else
-        = content.name
-    %dt Body
+        = content[:name]
+
+    %dt Short summary
     %dd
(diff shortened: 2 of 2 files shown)


app/models/revision_content.rb
```diff
@@ -19,6 +19,14 @@ def all_content
     end
   end
 
+  def keys_in_order
+    [:name, :content_body] + node.all_content.keys.reject do |key|
```
ltankey: Would it be worth us including `summary` and `short_summary` as well? They are now fields which are also a part of the base Node class.

tessereth:
I'm just thinking that the long summary is also markdown. So the name and short summary are plain text and the content body and long summary (aka 'summary') are markdown. It would also be nice to put them in the same order in the submission page as they are when you create a submission. But I'm not sure how you would create a simple way to specify that once and have it all be consistent.

freshtonic:
@tessereth addressed your comments.

app/models/revision_content.rb
```diff
@@ -19,6 +19,18 @@ def all_content
     end
   end
 
+  def main_keys_in_order
+    [:name, :content_body, :short_summary, :summary]
```
tessereth: I think this should be `[:name, :short_summary, :summary, :content_body]`

tessereth:
Apart from the order of the main keys, and a vague feeling that the template could be more DRY, this LGTM.

freshtonic:
@tessereth there are 3 different ways to render a key/value.

- as markdown
- as not markdown
- as not markdown but with a default.

So I didn't try to DRY it any further because of the permutations.

ltankey:
This seems good for revisable fields but doesn't work for metadata fields (ie `data`) which means that for News Article, release date wouldn't be shown in the submission review.

freshtonic:
@ltankey so if release date can change, then surely it should be revisable?

ltankey:
It can change, but for whatever reason it's been stored outside of content.

I guess the broader issue is that node creation of a combination of revisable changes and meta data changes, and we'll likely need to show all upon initial submission. This would be true of options, or any field unique to a node type.

ltankey:
@freshtonic from conversations with @tessereth - we'd need to relook at how we handle content creation (not update) and approvals, and that can happen outside of this PR. LGTM.

micapam:
I agree it should be revisable, but (a) it's not content, it's metadata, and (b) diff mechanism (LCS) we use for content doesn't apply for things like release date, we should use something like JSON-diff instead, so it shouldn't be a content attribute but handled differently.

PR #281, opened 2016-07-28
SITES-442 Static related topics for department home pages

micapam:
Static content for related topics. 
I've consolidated this with the other static stuff (categories on the homepage) into a single YAML file. 

n.b. This is only a WIP because I'm waiting for three out of four blurbs from content authors. Functionally it's complete.


ltankey:
LGTM

app/views/templates/_related_topics.html.haml
```diff
@@ -0,0 +1,16 @@
+- topic_data = STATIC_DATA[:department_topics][node.slug]
+- if topic_data.present?
+  %h2 Information and services
+  %p
+    = topic_data.blurb
```
tessereth: Should the blurb be in an "abstract" class?

tessereth:
Apart from possible abstract classes this LGTM.

tessereth:
update changelog post 0.7.1 release

micapam:
Changelog updated. 
Doesn't need abstract - this goes below the abstract.

PR #285, opened 2016-07-28
SITES-438 update changelog

freshtonic:


CHANGELOG.md
```diff
@@ -7,6 +7,8 @@ This file is influenced by http://keepachangelog.com/.
 
 ### Changed
 - Wrapped news summary in abstract block
+- Submissions now show all content fields other than just the name & body.
+  Meta-fields are still not displayed, however.
```
tessereth: Possibly don't need the clarification about the meta-fields. And then you could remove the full stop and it would stylistically match the rest of the changelog.

ltankey:
LGTM with the caveat that the meta-data comment probably isn't needed.

PR #286, opened 2016-07-28
[SITES-472] Authorization fixes

tessereth:
This includes refactoring the auth tests so it's a little easier to extend for new routes.

### DIFF
--- a/app/controllers/editorial/news_controller.rb
+++ b/app/controllers/editorial/news_controller.rb
@@ -2,6 +2,7 @@ module Editorial
   class NewsController < EditorialController
     include ::NodesHelper
     before_action :set_sections_by_membership, only: [:index, :new, :create, :edit]
+    before_action ->() { authorize! :create, :news }, only: [:new, :create]
 
     def index
       # TODO: return news editorial index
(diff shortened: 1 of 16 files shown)


app/controllers/editorial/news_controller.rb
```diff
@@ -12,10 +12,11 @@ def show
         section: params[:section],
         slug: params[:slug]
       )
-
+      authorize! :read, @article
     end
 
     def new
+      authorize! :create, :news
```
ltankey: Would it be more DRY to use a `before` hook?

app/controllers/editorial/submissions_controller.rb
```diff
@@ -19,6 +19,7 @@ def index
     end
 
     def create
+      authorize! :update, @node
```
ltankey: Again, RE: DRYing this up, perhaps we could use a before call.

tessereth:
update changelog post 0.7.1 release

ltankey:
These tests are much better - we're missing the `:owner` role and requests list but otherwise this is a great improvement.

LGTM

tessereth:
@ltankey There is no request list page. Or rather the route is there but it displays the editorial index page (probably a byproduct of our controller hierarchy). We should really remove that route until it's implemented.

micapam:
LGTM. Nice use of rspec shared examples

tessereth:
Updated to include topic owner and the request show route in the tests. Also update routes.rb not to route to request pages that don't exist.

PR #221, opened 2016-07-21
[SITES-364] Feature/view news from slash news

ltankey:
Provides an aggregation of all news items on `/news`.

image

PR #288, opened 2016-07-28
Refactor inline image usage for UI Kit v1.4.0

maxious:
We will now inline the images into CSS files locally rather than relying on getting them in a seperate css file. 

We do need the images in a zip file though so the rake task won't work completely until https://github.com/AusDTO/gov-au-ui-kit/pull/221 is merged and released (old icons should remain though)

(In pulling in v1.4 final, some minor fixes to homepage link styling and link markup were required)


ltankey:
LGTM

simonschwartz:
LGTM

maxious:
fixed homepage

PR #289, opened 2016-07-28
SITES-475 Validate news article release date

bjgordon:
Also fixed the news article release_date in seeds.

### DIFF
--- a/db/seeds.rb
+++ b/db/seeds.rb
@@ -23,6 +23,7 @@ news1 = NewsArticle.with_name(
     }
   ) do |news_article|
     news_article.name = "Business News"
+    news_article.release_date = Date.today
   end
 news1.revise!(content_body: 'foobar').apply!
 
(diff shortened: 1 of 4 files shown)


db/seeds.rb
```diff
@@ -23,6 +23,7 @@
     }
   ) do |news_article|
     news_article.name = "Business News"
+    news_article.release_date = Date.today
```
tessereth: Probably best to make this an update call afterwards because this block is only called if the article doesn't exist. We want to fix the release date for anyone who's already run seeds.

tessereth:
LGTM

tessereth:
Update changelog post 0.7.1 release

ltankey:
Just need to update changelog.

We will eventually need a plan to allow for drafts, but that's outside the scope of this PR.

bjgordon:
Updated CHANGELOG

ltankey:
LGTM

PR #301, opened 2016-07-29
[SITES-480] Be robust to node subclass submission forms not existing

tessereth:


ltankey:
LGTM

app/helpers/editorial_helper.rb
```diff
@@ -1,7 +1,9 @@
 module EditorialHelper
 
   def form_type(node_clazz)
-    "#{node_clazz.name}Form".constantize
+    # Because of development mode autoloading, the class may not exist until
+    # we call constantize. Hence just catching the exception.
+    "#{node_clazz.name}Form".constantize rescue NodeForm
```
micapam: I'm a bit dubious about this `rescue` for the sake of dev env....

micapam: I'm not quite sure of the scenario in which this could hide real problems but it feels like it might do.

ltankey: Would it be better to be explicit? We could make a `CustomTemplateNodeForm` and have that extend `NodeForm`.

tessereth: We can create the subclass. I just don't want to because there's nothing to put in it and I can see this happening again in the future. I can specifically rescue from the correct exception if that makes it any better. I also didn't actually test that the is-this-constant-defined worked or not in dev. I'll try that too.

tessereth: So the internet was correct and using `const_defined?` doesn't work in dev. I've updated to only rescue from NameError so it shouldn't be hiding anything major.

micapam:
LGTM

PR #306, opened 2016-07-29
SITES-450 demo script

freshtonic:


tessereth:
please update changelog post 0.8.1 release

spec/database_dump_loader.rb
```diff
@@ -0,0 +1,52 @@
+
+# Loads a Postgres database dump.
+#
+# Only COPY blocks are extracted. Any other commands, comments, whitespace is ignored.
+#
+# Skips content_blocks (because it's still in prod, and schema_migrations
+# (because db:test:prepare will already have popuated it and you'll get unique
```
coderdan: Minor typo

spec/database_dump_loader.rb
```diff
@@ -0,0 +1,52 @@
...
+# key violations).
+#
+class DatabaseDumpLoader
```
coderdan: This is nice!

spec/database_dump_loader.rb
freshtonic: ah yes

spec/features/demo_spec.rb
```diff
@@ -0,0 +1,58 @@
...
+
+    expect(page.current_path).to eql("/")
+    expect(page.title).to eql("GovAuBeta")
+    expect(page).to have_css('h2', :text => 'Information and services')
+    expect(page).to have_css('h2', :text => 'About Australia')
+    visit(page.find('.next-holiday a')[:href])
+
+    expect(page.current_path).to eql("/times-and-dates/australian-public-holidays")
+    expect(page).to have_css('h1', :text => 'Australian Public holidays')
+    visit(page.find('a', :text => 'School holidays and term dates')[:href])
```
coderdan: Is there a reason you're doing this instead of click? Does it even matter?

coderdan:
One small query but otherwise LGTM

spec/features/demo_spec.rb
```diff
@@ -0,0 +1,58 @@
...
+  end
+
+  it "should be possible to demo the application" do
+    visit '/'
+
+    expect(page.current_path).to eql("/")
+    expect(page.title).to eql("GovAuBeta")
+    expect(page).to have_css('h2', :text => 'Information and services')
+    expect(page).to have_css('h2', :text => 'About Australia')
+    visit(page.find('.next-holiday a')[:href])
```
micapam: This isn't a problem with the test per se, but it exposes an accessibility problem... should be able to use something other than CSS class.

spec/features/demo_spec.rb
freshtonic: @coderdan does `click_link` take a selector? There is one case in the demo where there are two links with same text on the page.

coderdan: Not sure. Looking it up now...

spec/features/demo_spec.rb
```diff
@@ -0,0 +1,58 @@
...
+    visit(page.find('article.content-listing li a', :text => 'Department of Communications and the Arts')[:href])
+
+    expect(page.current_path).to eql('/department-of-communications-and-the-arts')
+    visit(page.find('aside li a', :text => 'Our role')[:href])
+
+    expect(page.current_path).to eql('/department-of-communications-and-the-arts/our-role')
+    visit(page.find('aside li a', :text => 'Careers')[:href])
+
+    expect(page.current_path).to eql('/department-of-communications-and-the-arts/careers')
+    visit(page.find('aside li a', :text => 'Our role')[:href])
```
micapam: I think these precise selectors make the test too closely tied to the implementation = too fragile. Should only go to that level of detail if there's actual ambiguity on the page (e.g. more than one 'Our role' link).

freshtonic: There is a case in the demo where two links with the same text appear on the page (dept of comms is in the footer can can appear in main content, so need to disambiguate).

spec/features/demo_spec.rb
coderdan: You might able to do something like:

    page.find('a', :text => 'School holidays and term dates').click

See http://www.rubydoc.info/github/jnicklas/capybara/master/Capybara/Node/Element#click-instance_method

spec/features/demo_spec.rb
freshtonic: I think `aside li a` can be `aside a`. The demo google doc *specifically* says "click a link in the sidebar", so I think that's reasonable.

micapam: Yeah - that sounds like a good balance between disambiguation & the semantics > syntax principle.

freshtonic:
@coderdan so regarding `click_link` versus `visit(page.find('aside li a', :text => 'Our role')[:href])`

coderdan:
Honestly @freshtonic I don't think its an issue. I was just curious to know the motivation. As I said: LGTM :)

micapam:
LGTM !

PR #317, opened 2016-07-30
[SITES-503 | SITES-474] Feature/news section navbar

ltankey:
This PR introduces the navbar for section news index, and sets news item headings to be h3 instead of h2s. 

Although this is completely wrong from a structural and accessibility perspective, it has been requested to fit an appropriate look. Furthermore, a future h2 is expected to be added for refining the filtering process, which _may_ make these h3s valid, although the structural considerations are best addressed at that point.

This PR also fixes the navbar to indicate the `current` page, as well as including the section header at the top of the list.


micapam:
Would it be possible to keep them as `h2`s but apply a CSS class to make them look like `h3`s? 

Just a thought - otherwise LGTM.

ltankey:
Potentially @micapam but the h3 stylings both are the correct markup and have the correct styling. It might be easiest if we leave it without a h2 temporarily until the page has a filter widget.

app/views/application/_navbar.html.haml
```diff
@@ -1,11 +1,15 @@
 -# locals: menu_nodes, active
 
 %aside.sidebar
-  %nav.primary-nav{'aria-label' => 'main navigation'}
+  %nav.local-nav{'aria-label' => 'main navigation'}
     %a#primary-nav-heading.js-accordion-trigger.nav-heading{href: 'javascript:void(0)',
             'aria-expanded' => true, 'aria-controls' => 'primary-nav-panel'}
```
tessereth: Should it be #local-nav-heading now that the class has changed to local-nav?

tessereth:
Aside from minor comment, LGTM

app/views/application/_navbar.html.haml
ltankey: Potentially - I intentionally left the JS stuff as I'm not sure what their implementation is relying on. I'll make the change because it makes sense to, and consult with the uikit folk.

PR #319, opened 2016-07-30
Include incumbent minister's name next to related minister links - SITES-269

micapam:
This change uses the 'summary' field of a Minister section to store the incumbent minister's name, and reveal it alongside the links to related ministers in department homepages. 

It also removes the display of the 'summary' field from the public-facing header on section home nodes, so that the incumbent minister's name does not end up being shown there (it's already in the content for each under 'About the Minister'). This shouldn't be a problem - I have checked every single section home page on prod and the only one with a populated 'summary' field for that section (and thus appearing in the header) is TV reception - which could simply use the section home node's abstract field (which is currently empty) instead.

I am aware this is not an ideal solution. However, I don't think it's possible to implement an ideal or even good solution (using some kind of dynamic content block population as discussed in recent tech strategy meeting) in time for the demo this coming week. 

The only other alternative would be to use static content, as per the related topics. But the only reason we used static content there was because most topics are not yet created on prod. The ministers all seem to exist on prod. Plus building that static data would be a waste of time whereas this allows us to use the actual relations between Department and Minister sections - the only data debt would be to  shift the Minister summaries into whatever dynamic content solution we end up creating in the medium term.

This is the best I can come up with in the time (but happy to hear alternative suggestions, of course).

n.b. This resolves SITES-269


maadonna:
Not sure if you know this, but I dropped all the human names into the content model spreadsheet (the same one that has the list of Departments, Agencies etc). Might help you to grab the content automatically, or you might have already done this :)

ltankey:
LGTM

freshtonic:
LGTM

PR #329, opened 2016-08-01
SITES-511 Use h3s instead of p.abstract in dates and times custom templates

micapam:
As per request in ticket. 
Note that the description on SITES-511 was confusing, I clarified the requirements in Slack.


ltankey:
Where are the h2s?

micapam:
@ltankey Good question.

micapam:
My thinking: these pages are temporary anyway. I could make a `h2` with a class to style like a `h3`, instead, if you think it's worthwhile ?

ltankey:
I'm not sure - I don't want to create overhead for temporary pages, but I think we should be more concerned if we're getting invalid markup to match stylings. Would the content folk be able to offer up a valid h2 for the document structure so that the h3s are styled correctly and valid?

If this is blocking I don't mind offering a LGTM.

simonschwartz:
Could we just restyle the current abstract on times and dates pages to match the required visual? The `<h2>` actually exists further down the page within the callout 'The next public holiday is:'.

micapam:
Sounds sensible @simonschwartz - maybe `p.psuedo-h3` or something ?

simonschwartz:
How about something a little more semantic like '<p class="abstract--dates">'?

ltankey:
LGTM

app/assets/stylesheets/_times-dates-list.scss
```diff
@@ -36,3 +36,7 @@
     }
   }
 }
+
+p.abstract--dates {
+  @extend h3;
```
simonschwartz: Being picky, but I would recommend avoiding `@extend` as we found it created bugs in the UI Kit. It might be best to simply copy/paste the scss for h3. 
CC: @klepas 

https://www.sitepoint.com/avoid-sass-extend/

micapam: When I looked up the source for `h3` in UIKit I found it extends `h4` :)

simonschwartz:
LGTM

maxious:
LGTM

PR #332, opened 2016-08-01
[SITES-512] Feature/aggregate news by distribution

ltankey:
This PR ensures that a `/:section/news` block includes any news item for which the section is marked as a distribution.


app/controllers/editorial/news_controller.rb
```diff
@@ -35,10 +35,13 @@ def create
           # TODO: move this into a form validator
           # Prevents distribution list being created for publisher
```
tessereth: Is this comment now incorrect?

tessereth:
LGTM

micapam:
So this duplicates the `section` into `sections` collection - is that right ?

ltankey:
Correct. Originally I had attempted to combine two scopes but was getting grief getting the query from the `Section` working, as it would have to apply an `or` statement on both published and distribution relationships.

It's far simpler to treat all sections as the same for distributions, and restrict what gets rendered where in the views.

maxious:
LGTM but Merge conflicts

PR #334, opened 2016-08-01
SITES-508 send confirmation emails

freshtonic:
Production & staging require the following environment variables:

(NOTE: still waiting on production AWS env vars to be configured).

- `AWS_ACCESS_KEY`
- `AWS_SECRET_KEY`
- `FROM_EMAIL`
- `APP_DOMAIN`


freshtonic:
The `letter_opener` gem opens confirmation emails in your browser in the development environment.

It's worth noting that the confirmation email text is simply the Devise default right now.

It looks like this:

image

config/environments/development.rb
```diff
@@ -33,6 +33,8 @@
 
   config.action_mailer.default_url_options = { host: 'localhost', port: 3000 }
 
+  config.action_mailer.delivery_method = :letter_opener
```
tessereth: When I was doing request for membership emails, I set up mailtrap. See `config/initializers/mailer.rb` and the readme. How does this interact with mailtrap? Would this replace it?

freshtonic: @tessereth yes, I noticed after the fact.

So letteropener runs entirely locally and when an email is sent in in the development env it opens a browser tab containing the email.

I'm open to using mailtrap but I would rather not depend on a 3rd party service unless there's other compelling reasons to do so.

config/initializers/amazon_ses.rb
```diff
@@ -0,0 +1,7 @@
+
+if Rails.env.production?
+  ActionMailer::Base.add_delivery_method(:ses, AWS::SES::Base,
+    access_key_id:      ENV['AMAZON_ACCESS_KEY'],
+    secret_access_key:  ENV['AMAZON_SECRET_KEY']
```
ltankey: Do we bother having defaults if the environment doesn't provide these details?

db/schema.rb
```diff
@@ -10,11 +10,19 @@
...
   enable_extension "plpgsql"
 
+  create_table "content_blocks", force: :cascade do |t|
+    t.integer  "node_id"
+    t.text     "body"
+    t.datetime "created_at", null: false
+    t.datetime "updated_at", null: false
+    t.index ["node_id"], name: "index_content_blocks_on_node_id", using: :btree
+  end
+
```
ltankey: Please kill this with fire.

freshtonic: OH FFS it's like an unkillable whack-a-mole @ltankey

ltankey:
LGTM

config/environments/production.rb
```diff
@@ -64,6 +64,13 @@
   # Set this to true and configure the email server for immediate delivery to raise delivery errors.
   # config.action_mailer.raise_delivery_errors = false
 
+  config.action_mailer.delivery_method = :ses
+
+  config.action_mailer.default_url_options = {
+    protocol: 'https',
+    host: 'gov-au-beta.apps.platform.digital.gov.au'
```
tessereth: This should come from environment as this is not 12 factor

freshtonic: @tessereth yes it should - do we already have an env var for that?

ltankey: @freshtonic you can access them via `VCAP_APPLICATION` environment variable which should contain JSON we need to parse (https://docs.cloudfoundry.org/devguide/deploy-apps/environment-variable.html#app-system-env)

We'd need to access `application-uris` and regex out any coloured URLs (-blue and -green variants). The remaining uri would be the one we want to keep. 

**note** during our deploy, we're likely to remove the route to our app (which would leave _only_ the coloured variant URI). As such, we'd need to use regex to remove the colour from the URI we have access to.

config/initializers/amazon_ses.rb
freshtonic: What should happen is that the app refuses to start and produces a helpful error message detailing all missing configuration.

But nothing else in the app does that or provides defaults.

We should revisit later and flag as tech debt.

db/migrate/20160801053501_add_confirmed_at_to_users.rb
```diff
@@ -0,0 +1,8 @@
+class AddConfirmedAtToUsers < ActiveRecord::Migration[5.0]
```
maxious: What happens if an account existed before these columns were created? Will they be able to log in?

freshtonic: Amazing catch - no they won't!

Aren't code reviews _awesome_!

freshtonic: @maxious I can fix it in the DB but it does mean that we will *never* send confirmation emails for existing users. Are we OK with that?

maxious: I think so. Maybe include a date created filter before today or something?

tessereth:
Update administrate with the new user columns?

tessereth:
LGTM pending testing in staging

PR #343, opened 2016-08-02
Make agencies link null again - SITES-483

micapam:
This change got overwritten in the mêlée yesterday. Restoring it.


tessereth:
How did that happen? Odd. LGTM.

ltankey:
LGTM, would review again

Issue #299, opened 2016-07-29
Group/"Popular" links breakpoints not reacting as expected


hannah-ustwo:
Links should be appearing across 4 columns up until the mobile breakpoint where it changes to 2 columns.

Double checked and confirmed that it is working in the UI kit; appears to only be happening on the gov.au beta.

![e219af512cb3bd5c4cf386455d4ba5e3](https://cloud.githubusercontent.com/assets/19661064/17234119/21f29acc-5579-11e6-8347-a40e55690284.gif)

ltankey:
Thanks @hannah-ustwo - the markup seems identical from the example provided on the ui-kit page.

Just to confirm, this is correct:

```
         <div class='links-group'>
            <h2>Popular links</h2>
            <ul class='popular-now'>
              <li>
                <a href="/times-and-dates/australian-public-holidays">Public Holiday dates</a>
              </li>
              <li>
                <a href="#">Lodge your tax return</a>
              </li>
              <li>
                <a href="#">Working in Australia</a>
              </li>
              <li>
                <a href="#">Visiting Australia</a>
              </li>
              <li>
                <a href="#">What can you bring to Australia</a>
              </li>
              <li>
                <a href="#">Citizenship</a>
              </li>
              <li>
                <a href="/times-and-dates/school-holidays-and-term-dates">School Holiday dates</a>
              </li>
              <li>
                <a href="#">Government Jobs</a>
              </li>
            </ul>
          </div>
```

hannah-ustwo:
The markup looks right...any other clues to what might be making the breakpoints malfunction @klepas @petronbot ?

ltankey:
@hannah-ustwo do you have any full-page examples that we might be able to look at? Is it possible some of the header changes have had flow-on effects? We are now using `prerelease-govau--header`.

hannah-ustwo:
@ltankey Not sure if this is helpful, but we have an example page before we introduced the `prerelease-govau--header` [over here](http://gov-au-ui-kit.apps.staging.digital.gov.au/examples/group-links.html)

Looping in @treb, he's looking into it now.

treb:
?????????

TrebBrennan:
I'm guess you meant me. Sorry @treb please disregard

hannah-ustwo:
Ooops...

streetdaddy:
triaged as SITES-496

treb:
?????????why are you sending me emails?

[...]

PR #355, opened 2016-08-02
SITES-496 Bugfix/popular breakpoints

bjgordon:
Fix for #299 
Not sure why this works exactly, but rails-sass seems to need the arguments this way to work. Will also put in a PR to ui-kit to fix upstream.


tessereth:
changelog? Otherwise LGTM.

maxious:
LGTM

PR #360, opened 2016-08-02
SITES-521 whitelist email domains to *.gov.au

freshtonic:


maxious:
LGTM
(the only exception I know of is csiro.au but later problem)

ltankey:
Will this fail validation on admin?

freshtonic:
@ltankey yes - and that means we'll need to change the admin login.

app/validators/email_domain_whitelist_validator.rb
```diff
@@ -0,0 +1,7 @@
+class EmailDomainWhitelistValidator < ActiveModel::EachValidator
+
+  def validate_each(record, attribute, value)
+    record.errors.add(attribute, 'only *.gov.au email addresses permitted') if value !~ /\.gov\.au$/
```
tessereth: Should it be `\z` instead of `$`?

freshtonic: Yes it should.

freshtonic: Fixed.

ltankey:
LGTM

app/validators/email_domain_whitelist_validator.rb
micapam: Out of interest - what's the difference? I always use `$`.

PR #365, opened 2016-08-02
SITES-192 set AWS SES server in correct region

freshtonic:


config/initializers/amazon_ses.rb
```diff
@@ -2,6 +2,7 @@
 if Rails.env.production?
   ActionMailer::Base.add_delivery_method(:ses, AWS::SES::Base,
     access_key_id:      ENV['AWS_ACCESS_KEY'],
-    secret_access_key:  ENV['AWS_SECRET_KEY']
+    secret_access_key:  ENV['AWS_SECRET_KEY'],
+    server: "email.us-west-2.amazonaws.com"
```
ltankey: This really feels like it should be an environment variable.

ltankey:
LGTM

maxious:
LGTM

PR #367, opened 2016-08-03
[SITES-570] add transition links to root

ltankey:
image


app/views/templates/root_node.html.haml
```diff
@@ -32,6 +32,28 @@
   %a.see-more More about Australia
 
   %section.home-news-list
```
tessereth: Should it be styled with the home-news-list class or something else?

ltankey: Nope

tessereth:
LGTM

bjgordon:
LGTM, but amend the commit message.

PR #373, opened 2016-08-04
[SITES-343] Feature/commit and tag in footer

ltankey:
Adds the deployed tag/sha1 to the editorial footer if one is present.

This is supplied by our deploy process - prod will get both the git tag and sha1, whereas staging will only get the sha1.

### DIFF
--- a/bin/deploy-prod.sh
+++ b/bin/deploy-prod.sh
@@ -9,6 +9,12 @@ set -o pipefail
 # echo out each line of the shell as it executes
 set -x
 
+# Update the circle envvars for all apps
+cf set-env gov-au-beta-blue CIRCLE_TAG "$CIRCLE_TAG"
+cf set-env gov-au-beta-blue CIRCLE_SHA1 "$CIRCLE_SHA1"
+cf set-env gov-au-beta-green CIRCLE_TAG "$CIRCLE_TAG"
+cf set-env gov-au-beta-green CIRCLE_SHA1 "$CIRCLE_SHA1"
+
 # Update the blue app
 cf unmap-route gov-au-beta-blue apps.platform.digital.gov.au -n gov-au-beta
 cf push gov-au-beta-blue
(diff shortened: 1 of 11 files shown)


tessereth:
tests?

ltankey:
Added a feature test for `editorial`

spec/features/editorial_version_spec.rb
```diff
@@ -0,0 +1,22 @@
+require 'rails_helper'
+
+RSpec.describe 'editorial version details', type: :feature do
+  Warden.test_mode!
+
+  let!(:root_node) { Fabricate(:root_node) }
```
tessereth: The `:root_node` fabricator should not be required. It should be created by the section fabricator.

config/environments/test.rb
```diff
@@ -46,4 +46,8 @@
   # config.action_view.raise_on_missing_translations = true
   config.active_job.queue_adapter     = :async
   config.active_job.queue_name_prefix = "gov-au-beta_#{Rails.env}"
+
+  # Set mock environment git tag and SHA1 values
+  config.version_tag = 'v1.0.0'
+  config.version_sha = 'abcdefghijkl'
```
tessereth: Do we want to set some dummy values in dev as well? Just so we can see what it would look like once it gets to prod?

bin/deploy-prod.sh
```diff
@@ -9,6 +9,12 @@ set -o pipefail
 # echo out each line of the shell as it executes
 set -x
 
+# Update the circle envvars for all apps
+cf set-env gov-au-beta-blue CIRCLE_TAG "$CIRCLE_TAG"
+cf set-env gov-au-beta-blue CIRCLE_SHA1 "$CIRCLE_SHA1"
+cf set-env gov-au-beta-green CIRCLE_TAG "$CIRCLE_TAG"
+cf set-env gov-au-beta-green CIRCLE_SHA1 "$CIRCLE_SHA1"
+
```
bjgordon: Any reason why you're setting these before rather than after the deploy has run?

bjgordon: Oh wait, the app won't update to show the new vars until restarted, so is ok to set before.

bjgordon:
I think it'd be nice to show a timestamp as well to help non-devs, but LGTM.

maxious:
LGTM - will see how goes on staging ;)

PR #374, opened 2016-08-04
SITES-515 The html title tag should reflect the page name

maxious:
Edited all the templates including times/dates and editorial to copy from the H1.


tessereth:
This works but... I feel like there must be a better way. Unfortunately, I can't think what it is. Maybe we could use `yield :title` inside the h1 so at least the h1 and title would always be the same?

ltankey:
We should consider how we will manage this with content - should the H1 of a node's content inform the title? If so, the interaction may be a little more complex. If we're to set the title based off the content, we should have a test for this.

ltankey:
After further thought, I don't think a user should be setting a H1 for the most part - it is likely to come from a page template or node name. My concerns could be better addressed in a content analysis card.

ltankey:
LGTM

tessereth:
Given I can't think of a better way to do this, LGTM.

app/views/layouts/editorial.html.haml
```diff
@@ -1,7 +1,7 @@
 !!! 5
 
 %html.js
-  = render 'head', title: 'GovAuBeta: Collaboration'
+  = render 'head', title: (content_for?(:title) ? "#{yield(:title)} | GOV.AU Collaboration" : 'GOV.AU Collaboration' )
```
bjgordon: I know it was previously Collaboration, but should the title just be Editorial? This is the only place we call it Collaboration. The breadcrumbs, title etc is all Editorial.

maxious:
rebased

PR #375, opened 2016-08-04
[SITES-542] Whitelist types when creating a node

tessereth:
Programmers' changelog:
* Move the check for valid node types above the constantize call
* Return a 400 error when there's a bad type
* Remove some deprecation warnings in tests
* Remove duplicated tests


ltankey:
LGTM

bjgordon:
LGTM

PR #380, opened 2016-08-04
[SITES-343] Feature/commit and tag in footer

ltankey:
Because it'll be rejected in QA

* improved clarity around the version blob in the footer
* truncated hash

image

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,6 @@ This file is influenced by http://keepachangelog.com/.
 - Include git tag and sha1 on editorial footer
 
 ### Changed
-- Truncated SHA on editorial page and improved clarity around version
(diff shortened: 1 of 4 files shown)


CHANGELOG.md
```diff
@@ -7,6 +7,7 @@ This file is influenced by http://keepachangelog.com/.
 - Include git tag and sha1 on editorial footer
 
 ### Changed
+- Truncated SHA on editorial page and improved clarity around version
```
tessereth: This is probably not needed because it's be covered by the previous changelog update that is in the same version.

tessereth:
Apart from possible changelog changes, LGTM.

bjgordon:
LGTM

PR #383, opened 2016-08-05
SITES-555 submission template per node type

freshtonic:
When implementing this feature I noticed that out node specs were instantiating the Node base class everywhere. I fixed this, but it's a large PR (sorry!). See my commit messages for more details.


app/controllers/editorial/submissions_controller.rb
```diff
@@ -9,6 +9,12 @@ class Editorial::SubmissionsController < Editorial::SectionsController
     before_action :find_submission, only: [:show, :update]
     decorates_assigned :submission
 
+    def show
+      super
+      # FIXME: this works around the node Fabricator not setting the type in specs
```
freshtonic: I'll fix this in the Fabricator

tessereth: You can possibly use `#class` rather than `#type` if `#type` has problems.

freshtonic: The issue is not that Fabricator doesn't support `type`. It's that loads of our specs need fixing - the majority are written against `Node`, not `Node` subclasses. When I set the type to `GeneralContent` by default in the Fabricator, loads of specs break.

tessereth:
So when I was refactoring the submission show page, my thought was that non-revisable things would go under the 'metadata' section and revisable things in the 'original content' and 'submitted content' sections. I'm not saying this is necessarily correct, but would it make more sense for the release date of a news article to be in the metadata section because it's currently not revised? Also, did you test this for news articles? I didn't think RevisionContent['date'] was the way that worked but my knowledge is somewhat out of date.

app/controllers/editorial/submissions_controller.rb
freshtonic: (which I will fix this morning BTW :) )

tessereth: Ew. Okay then.

freshtonic: Yeah it's not a fun job but tests that create instances of the base class aren't testing things production does in Real Life™ 😜

spec/controllers/editorial/submissions_controller_spec.rb
```diff
@@ -156,7 +160,14 @@
   end
 
   context 'as a reviewer' do
-    let!(:submission) { Fabricate(:submission, state: :submitted) }
+    let!(:submission) { 
+      begin
+        Fabricate(:submission, state: :submitted)
+      rescue
+        binding.pry
```
maxious: Is this still required?

freshtonic: LOL, good catch

app/views/editorial/submissions/node_types/_news_article.html.haml
```diff
@@ -0,0 +1,3 @@
+
+%dt Release date
+%dd= content["date"]
```
bjgordon: I dont think this is right - it should be `release_date`, and it's also not revisable so I dont think it would be in revision_content? I think we need to get `release_date` from the NewsArticle node.

app/views/editorial/submissions/_content.haml
```diff
@@ -34,10 +34,4 @@
...
           != markdown_content content[:content_body]
 
-    - content.optional_keys_in_order.each do |key|
-      %dt= key.to_s.titleize
-      %dd
-        - if content[key].nil?
-          none
-        - else
-          = content[key]
+    = render partial: "editorial/submissions/node_types/#{node_type_template}", locals: { content: content }
```
bjgordon: Maybe it would be cleaner if rather than requiring us to make an (empty) file for each node type, (i.e. 3/5 of them are empty at the moment) we only call render partial if `node_type_template` exists?

freshtonic: I disagree - it's simpler as it is. Everything follows the pattern. No exceptions.

app/views/editorial/submissions/node_types/_news_article.html.haml
freshtonic: You're correct - this is wrong.

However, it means that I cannot show the release date for a node here because it's not part of the revision. I think this is highlighting the fact that we have an arbitrary mix of versioned and non-versioned fields.

I vote that for now we just delete the content of this file, while we have a rethink of how we tackle versioning.

freshtonic:
Closing because versioning of fields needs more thought.

PR #385, opened 2016-08-08
SITES-515 Fix double title for nodes in editorial

maxious:
As per review feedback on SITES-515.
Also fix name of Editorial section.


app/views/editorial/nodes/show.html.haml
```diff
@@ -1,6 +1,3 @@
-= content_for :title do
-  = node.name
-
```
tessereth: Maybe you could add a comment explaining why this file doesn't have a `content_for :title`? Then if someone comes by later and notices it's not there, they won't add it by accident. And it we change the way previews are handled, we know we might need to add it back.

tessereth:
Other than one minor thing, LGTM.

bjgordon:
LGTM

maxious:
rebase'd

PR #388, opened 2016-08-09
SITES-551 Add rake task to dump the database to file and upload to S3

bjgordon:
I have also changed the AWS variable names used in configuring SES so they match those used in the aws-sdk (AWS_ACCESS_KEY_ID and AWS_SECRET_ACCESS_KEY). I have already added the new names alongside the existing ones in staging and prod.


lib/tasks/db.rake
```diff
@@ -0,0 +1,49 @@
...
+    puts cmd
+    unless system cmd
+      raise 'Database dump failed'
+    end
+  end
+
+  desc 'Scrubs database dump file and uploads to S3'
+  task :s3upload => :dump do
+
+      scrubbed_file = "#{Rails.root}/tmp/db_scrubbed.sql"
```
tessereth: Should this use `backup_dir`?

bjgordon: Yep I'll change it.

lib/tasks/db.rake
```diff
@@ -0,0 +1,49 @@
+namespace :db do
+
+  desc 'Dumps the database to file'
+  task :dump => :environment do
+
+    host = ActiveRecord::Base.connection_config[:host]
+    db = ActiveRecord::Base.connection_config[:database]
+
+    cmd = "pg_dump --no-owner --no-acl -v -f #{dump_file} --no-password "
```
ltankey: Where do we define this?

ltankey: Also, if we add `-c` as a parameter to `pg_dump`, we won't have to do a destroy beforehand as it will drop before creating anyway.

bjgordon: dump_file is a private function further down, so it could shard between the two tasks. I'll add -c.

maxious:
Conflicts :(

bjgordon:
Updated with security review comments.

lib/tasks/db.rake
```diff
@@ -0,0 +1,46 @@
+namespace :db do
+
+  desc 'Dumps the database to file'
+  task :dump => :environment do
+
+    host = ActiveRecord::Base.connection_config[:host]
+    db = ActiveRecord::Base.connection_config[:database]
+
+    cmd = "pg_dump --no-owner --no-acl -v -c -f #{dump_file} --no-password "
+    cmd += ENV['DATABASE_URL'] || "--host #{host} --dbname #{db}"
```
ltankey: If a `DATABASE_URL` is not present, specify _just_ the `host` would not be sufficient to establish a connection. You would likely need to form a URL similar to the format of `DATABASE_URL`, which is:

`postgres://username:pass@host:port/dbname`

bjgordon: Inferring from `database.yml`, my thinking is that in CF environments, `DATABASE_URL` must be present or the app won't start. The fallback to `--host ...` --dbname` is used when we are in development i.e. a local db where this should be sufficient. I might be wrong? I should probably add a comment to the code explaining this.

ltankey: Ahh OK, that makes sense. Perhaps a rake command explicitly for testing might be a better way of splitting that out :)

ltankey:
CHANGELOG? I know it's not user-facing but it's still good to log these changes.

lib/tasks/db.rake
bjgordon: Thinking about it more, it's not really just for testing, maybe you would want to use the rake task to dump the database out if you had a local db...
The line in question is just about adapting `database.yml` and the current environment to pg_dump. I can re-write it so it's a bit clearer...

ltankey:
LGTM

maxious:
LGTM

bjgordon:
Squashed and rebased

PR #391, opened 2016-08-11
SITES-602 do not instantiate Node base class in fabricators

freshtonic:
Fabricators were creating nodes that were not representable in real
cases. For example, instantiating the Node base class directly,
attaching it to the root node and treating it like a general content
node.

Fixing this subsequently caused lots of spec breakage due to ancestory
depth validations when using the correct node types.


maxious:
LGTM

ltankey:
LGTM

bjgordon:
LGTM not that you need it.

PR #392, opened 2016-08-12
SITES-409 An admin can create a category

maxious:
Allows creation of Category objects which can be arranged in a hierarchy. At the moment we expect a three level hierarchy.

Allows relating a section (but intended for topics only) to one or more categories.

Changed capybara backend to poltergeist/phantomjs because Qt won't compile and the binary releases have deprecated QtWebKit


config/routes.rb
```diff
@@ -47,6 +47,7 @@
     resources :sections
     resources :submissions
     resources :topics
+    resources :categories
```
bjgordon: I just noticed the order of resources in `routes.rb` is the order they are shown in the admin side-bar. Let's keep it in alphabetical order.

app/dashboards/category_dashboard.rb
```diff
@@ -0,0 +1,58 @@
...
+  # Feel free to add, remove, or rearrange items.
+  COLLECTION_ATTRIBUTES = [
+      :id,
+      :name,
+      :parent
+  ].freeze
+
+  # SHOW_PAGE_ATTRIBUTES
+  # an array of attributes that will be displayed on the model's show page.
+  SHOW_PAGE_ATTRIBUTES = [
```
bjgordon: Handy to also show `:slug`. We also usually include `created_at` and `updated_at`.

app/views/categories/show.html.haml
```diff
@@ -0,0 +1,23 @@
...
+
+
+%article.content-main
+  - @category.children.each do |subcategory|
+    %h2= subcategory.name
+    - subcategory.children.each do |leaf_category|
+      %h4= leaf_category.name
+      %ul
+        - leaf_category.sections.each do |section|
+          - if section.home_node.state.draft? and not section.home_node.options.suppress_in_nav
```
bjgordon: Not sure if this logic is right.

app/dashboards/category_dashboard.rb
```diff
@@ -0,0 +1,58 @@
...
+  # Each different type represents an Administrate::Field object,
+  # which determines how the attribute is displayed
+  # on pages throughout the dashboard.
+  ATTRIBUTE_TYPES = {
+      id: Field::Number,
+      name: Field::String,
+      short_summary: Field::String,
+      summary: Field::String,
+      parent: Field::BelongsTo.with_options(class_name: "Category"),
+      children: Field::HasMany.with_options(class_name: "Category"),
```
bjgordon: Can we also show the sections?
e.g. add `sections: Field::HasMany`, and add `:sections` to `SHOW_PAGE_ATTRIBUTES` ?

bjgordon:
LGTM

simonschwartz:
LGTM

PR #393, opened 2016-08-15
Update UI Kit to v1.7.3

simonschwartz:
The PR updates the UI Kit to v1.7.3. It also refactors out custom css and markup in the platform for tabs components which has now been included to the UI Kit.


bjgordon:
The new tabs sizing looks a bit different, is that intentional? Think it's caused by the change to <h2> 
image

simonschwartz:
@bjgordon the tabs are a bit bigger because they are using the new UI Kit styling rather than the temporary custom css previously used

bjgordon:
Please add a message to CHANGELOG

maxious:
If this fixes the color contrast, accessibility_spec.rb should have color contrast re-enabled. 
There are 4 color contrast errors on 1.7.2.

simonschwartz:
@maxious I believe there is a colour contrast issue on the homepage also (custom css). Should I address that in the PR also?

simonschwartz:
@bjgordon spacing and size of tabs on times and dates page has been updated to prevent wrapping on desktop

TrebBrennan:
Design reviewed.

simonschwartz:
@maxious Let's hold off on re-enabling colour contrast test. This version of the UI Kit doesn't fix the colour contrast issues. There are also colour contrast issues on the homepage which will be addressed

maxious:
LGTM

bjgordon:
LGTM

PR #396, opened 2016-08-16
SITES-593 tool to make a static site snapshot

freshtonic:


config/environments/static.rb
```diff
@@ -0,0 +1,21 @@
...
+  config.public_file_server.enabled = true
+  config.assets.js_compressor = :uglifier
+  config.assets.debug = false
+
+  Rails.application.routes.default_url_options = {
+      protocol: 'http',
+      host: 'localhost:3001'
+  }
+
+  config.content_analysis_base_url = "https://dummy"
```
bjgordon: This line is also in `external_apis.rb`, can we delete it from here?

lib/spider.rb
```diff
@@ -0,0 +1,221 @@
...
+
+      unless depth >= @max_depth
+        non_asset_links(nodes).each do |link|
+          crawl(URI.join(uri, link), depth + 1) # rescue nil
+        end
+      end
+    end
+
+    def external?(uri)
+      uri.host && uri.host != @root_uri.host
```
bjgordon: /pedantic should be a case insensitive comparison

bjgordon:
Minor comments aside, LGTM.

config/environments/static.rb
freshtonic: yes this can be deleted

ltankey:
Should we let environment variables set the target ports? It may be simpler for us to provide the option in order to play nice with a CF environment, unless any potential CF deploy would rely on nginx acting as a reverse proxy?

maxious:
I saw that more as a way to test that the crawl worked correctly rather than any way to host it publically

freshtonic:
@ltankey I envisage this being executed on a developer laptop with a snapshot of production. Not sure we need that - and if we do we can add it later.

ltankey:
@freshtonic ah I see, makes sense then. LGTM

PR #398, opened 2016-08-16
SITES-599 Sanitize HTML in markdown preview, remove unused editors

maxious:
Using https://github.com/punkave/sanitize-html ensure that javascript based editor previews only contain valid and safe HTML.

editor.md and trumbowyg were being tested but should stop.


app/views/editorial/nodes/_editor_injection.haml
```diff
@@ -6,7 +6,22 @@
       var text = $('#node_content_body')[0].value,
         target = $('#editor-preview')[0],
         converter = new showdown.Converter({tables: true}),
-        html = converter.makeHtml(text);
+        // default filter settings from https://github.com/punkave/sanitize-html made explicit below
+        html = converter.makeHtml(sanitizeHtml(text), {
+        allowedTags: [ 'h3', 'h4', 'h5', 'h6', 'blockquote', 'p', 'a', 'ul', 'ol',
```
bjgordon: The app will render `h1`, `h2`, and `img` tags if they're used in published content, so I think we should include in `allowedTags` so the live preview is accurate. Unless having `img` leaves the security flaw we're trying to fix?

bjgordon:
May as well delete `_editor_links.haml`

bjgordon:
Since we're removing all editor selection things, also remove @editor initialisation from `Editorial::SubmissionsController#new` plus everything that breaks.

maxious:
okay moved it to a JS file and added some more allowed tags

bjgordon:
LGTM

ltankey:
LGTM

PR #401, opened 2016-08-16
SITES-578 SITES-609 More robust scrubbing

bjgordon:
No longer assumes column order, and raises error if scrubbing does not occur.
Also now scrub IPs, reset tokens, and confirmation tokens.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,12 +8,11 @@ This file is influenced by http://keepachangelog.com/.
 - Added collecting test metadata for CircleCI
 - Added ability to mark pages as placeholders that appear in navigation
 
-
 ### Changed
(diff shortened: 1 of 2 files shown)


ltankey:
#405 introduces a few additional fields which should 100% be scrubbed (any TOTP or OTP field as well as phone_number and unconfirmed_phone_number). Do we want to include this here, or after #405 lands with a subsequent PR?

bjgordon:
The scrub script will error out if it doesn't find the columns, so we can't add them now. I reckon a later PR, or even better it could be included in #405. i.e. as we add fields we should add scrubbing at the same time.

ltankey:
LGTM

maxious:
LGTM

bjgordon:
Rebased

PR #403, opened 2016-08-17
SITES-606 Add feedback button on all headers

maxious:
Includes current page url from the current request object


app/views/layouts/application.html.haml
```diff
@@ -28,8 +28,9 @@
             .brandbar.default
               .b1
               .b2
-          = #render partial: 'feedback_and_menu'
-
+          .feedback{tabindex:1}
+            %a.button--feedback{href:"https://qfrankp.au1.qualtrics.com/jfe/form/SV_5BgphbOHhi7qXe5?pageviewed="+request.url, role:"button"}
```
ltankey: Could we use an environment variable, just in case it changes?

ltankey:
Changelog?

simonschwartz:
LGTM

maxious:
rebase'd

app/views/layouts/application.html.haml
```diff
@@ -28,8 +28,9 @@
             .brandbar.default
               .b1
               .b2
-          = #render partial: 'feedback_and_menu'
-
+          .feedback{tabindex:1}
+            %a.button--feedback{href:ENV.fetch('FEEDBACK_URL','')+request.url, role:"button"}
```
bjgordon: There's no requirement given, but I wonder if FEEDBACK_URL is empty should we just not show it at all?

maxious: Did consider but then we wouldn't be testing that code everytime.

bjgordon: Hmm maybe this needs a test then?

ltankey:
LGTM

PR #405, opened 2016-08-17
[SITES-590] Feature/2fa

ltankey:
**COMMENTS WELCOME**

This is bigger than I intended, because it essentially became an epic.

The primary goal of this PR was to introduce 2FA to the application via a Devise extension called `two_factor_authentication`. However, as this gem only handles authentication (and not subsequent identity verification or 2FA setup), there are additional functions included to support this (that aren't particularly DRY just yet).

The functions include:

* sign-in with 2FA (TOTP or OTP)
* initialise 2FA for accounts to complete setup
* verify identity via 2FA enforced with before_action calls 

The components include:

- [x] SMSService (as the name suggests)
- [x] one-time-password service for generating one-time-passwords (and corresponding _sent_at) values, along with validation methods
- [x] two_factor_authentication extension for Devise
- [x] 2FA-verification protection of sensitive fields

Still to do:
- [ ] google authenticator setup (qr code generation)
- [ ] limit one-time-password attempts/retries


app/controllers/users/two_factor_verification_controller.rb
```diff
@@ -0,0 +1,57 @@
+module Users
+  class TwoFactorVerificationController < ApplicationController
+    before_action :check_direct_otp_expiry, only: [:show, :create]
+
+
+    def show
```
freshtonic: Should be a POST (create).

app/controllers/users/two_factor_verification_controller.rb
```diff
@@ -0,0 +1,57 @@
...
+      unless current_user.totp_enabled?
+        current_user.create_direct_otp
+        r = SMSService.new.send_two_factor_confirmation_code!(current_user)
+
+        # TODO: make this async and handle service failure
+        unless r.code == 202
+          raise 'There was an error sending the SMS'
```
freshtonic: Should probably return a 400 ?

app/controllers/users/two_factor_verification_controller.rb
```diff
@@ -0,0 +1,57 @@
...
+        end
+      end
+    end
+
+    def create
```
freshtonic: Oh but there's a create here. Maybe there should be a different endpoint for sending the SMS, above?

ltankey: This comes down to how we expect the SMS service to work. It's currently modelled off the `two_factor_authentication` gem which essentially sends it on `#show`.

This has an advantage in that it doesn't require the user have to click through to have an SMS sent to them - however the disadvantage is it doesn't provide the user with a warning that 2FA will be required, or give them an opportunity to back out of it.

We could choose to use the show action to display an informative screen outlining that verification is required via 2FA, and get the user to click through (post to create) to send the SMS. The subsequent call to submit the code could be an update action.

app/controllers/users/two_factor_verification_controller.rb
ltankey: We probably do - but do we need to handle this in any sophisticated manner? If the SMS API is (intermittently or otherwise) down, do we want a strategy of retrying? If not, we could probably settle for displaying an informative message of `The 2FA service is currently down`, but would need to test how helpful that is.

freshtonic: As we discussed on slack, fire the SMS request off in a new thread and ignore it if it fails to send, but indicate to the user that they can try again if they do not receive an SMS.

app/controllers/users/two_factor_verification_controller.rb
freshtonic: The problem with `show` is that it's `GET` with a side-effect. How are you making sure that link does not get crawled?

ltankey: Really good point. I guess it'd be stopped at the outer by needing to be signed in, but your point still stands.

freshtonic: If you do end up deciding to keep the GET, you can add `rel="nofollow"` on links.

freshtonic: And also add that route to `robots.txt`

app/controllers/application_controller.rb
```diff
@@ -15,17 +16,41 @@ class ApplicationController < ActionController::Base
...
+    #   User.includes(:roles).find(user.id)
+    # end
   end
 
   def decorated_current_user
     current_user.try(:decorate)
   end
 
+  #TODO: move this into a Warden after_authentication method or
+  # something likely to be executed _after_ TFA has been skipped
```
tessereth: This comment is confusing. Can you be clearer about what you mean?

app/controllers/application_controller.rb
```diff
@@ -15,17 +16,41 @@ class ApplicationController < ActionController::Base
...
+  def complete_two_factor_setup
+    if current_user && !current_user.account_verified && current_user.confirmed_at
+      # Users need to be able to verify their tfa code
+      #unless request.path.start_with?('/users')
+      if request.path.match(/\A\/users(?!\/edit).*\z/) == nil
```
tessereth: This regex is confusing. If there's only a few paths that _don't_ require a redirect, it may make more sense to just white list them rather than making assumptions about routing.

app/controllers/users/two_factor_setup_controller.rb
```diff
@@ -0,0 +1,55 @@
...
+
+    def create
+      current_user.unconfirmed_phone_number =  params[:phone_number]
+
+      if current_user.save
+        current_user.send_new_direct_otp_for(:unconfirmed_phone_number)
+        redirect_to confirm_users_two_factor_setup_path and return
+      end
+
+      render :new
```
tessereth: Stylistically, we've tended to put the `render :new` in an else clause rather than having an early return.

app/services/otp_service.rb
```diff
@@ -0,0 +1,60 @@
...
+  end
+
+
+  def create_otp(field)
+    create_direct_otp(field)
+  end
+
+
+  def authenticate_otp(code_field, code)
+    direct_otp = @user.send(code_field)
```
tessereth: May be worth adding a comment to this method "Never under any circumstances call with a user provided code_field argument" or something to that effect.

app/validators/aus_phone_number_validator.rb
```diff
@@ -0,0 +1,7 @@
+class AusPhoneNumberValidator < ActiveModel::EachValidator
+  def validate_each(record, attribute, value)
+    if value != nil && value !~ /\A04[0-9]{8}\z/
```
tessereth: This is a little pedantic. It would be much nicer to allow the user to put spaces in their number and just strip them server side.

ltankey: Agreed, although I think there's a whole flow around setting up phone numbers and authenticator details that's likely going to be left for another PR. I think we will address this validator concern as part of that.

app/views/devise/two_factor_authentication/show.html.haml
```diff
@@ -0,0 +1,28 @@
+%artcile.sign-in
```
tessereth: Misspelling of article?

tessereth:
Missing tests but otherwise LGTM

tessereth:
Update changelog post 1.1.0 release

CHANGELOG.md
```diff
@@ -8,6 +8,7 @@ This file is influenced by http://keepachangelog.com/.
 - Added collecting test metadata for CircleCI
 - Added ability to mark pages as placeholders that appear in navigation
 - Rake task for dumping, santising, and uploading the database to AWS S3
+- Added two-factor authentication to sign-in/up and verification
```
maxious: will need v1.1.0 rebase

app/views/users/two_factor_setup/confirm.html.haml
```diff
@@ -0,0 +1,23 @@
+%artcile.verify
```
maxious: Misspelling article might have bad effects on UI Kit styling.

bin/scrub.rb
```diff
@@ -33,6 +33,16 @@
       parts[col_index[:last_name]] = 'Nameless'
       parts[col_index[:confirmation_token]] = '\N'
       parts[col_index[:unconfirmed_email]] = '\N'
+      parts[col_index[:phone_number]] = '0412345678'
```
maxious: perhaps just REDACTED would be easier to read? Took me a sec to understand.

maxious:
Minor tweaks req. but LGTM

PR #407, opened 2016-08-18
SITES-619 generate ETAGs in all show/index requests

freshtonic:
The generated ETAG is automatically busted when a new version of the app
is released.

NOTE: does not cover admin routes.

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -1,3 +1,5 @@
+require 'digest/sha1'
+
 class ApplicationController < ActionController::Base
   # Prevent CSRF attacks by raising an exception.
   # For APIs, you may want to use :null_session instead.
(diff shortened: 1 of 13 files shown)


maxious:
Categories controller?

freshtonic:
@maxious done

maxious:
LGTM

app/controllers/application_controller.rb
```diff
@@ -17,12 +19,38 @@ def current_user
 
   protected
 
+  # Calls *stale?* but modifies any supplied *strong_etag* by prepending the value of 
+  # ApplicationController#etag_seed.
+  def bustable_stale?(object = nil, **kwd_args)
```
coderdan: I think one * is enough here, no? The double * is a c++ thing?

micapam: Array versus hash ... * for normal args, ** for named args

freshtonic: What Josh said :)

coderdan: Ah, right. Cool!

coderdan:
LGTM

freshtonic:
Waking up LGTM

PR #409, opened 2016-08-19
Fix use of placeholder flag for topics

maxious:
I added a flag for nodes but forgot to use them for section home nodes too. Oops.

This makes it so editorial can also control topic visibility without admin rights which is desired by users.


bjgordon:
LGTM

freshtonic:
LGTM

maxious:
Rebased

PR #410, opened 2016-08-19
SITES-618 improve editorial index performance

freshtonic:
- Loading the Section now preloads the home_node for the Section
- ApplicationController#current_user now preloads all of the User roles

Also added a feature to ensure that Section#name is updated when
SectionHome#name is updated. Ideally we can restore the delegate at some
point but Sections are created *before* the SectionHome in the admin
area and they will need to display a name in the meantime.

Fixed related breakages.

### DIFF
--- a/app/decorators/user_decorator.rb
+++ b/app/decorators/user_decorator.rb
@@ -9,20 +9,8 @@ class UserDecorator < Draper::Decorator
     end
   end
 
-  def pending_request_for(section)
-    object.requests.find_by(section: section, state: 'requested')
-  end
 
-  def pending_request_for?(section)
-    pending_request_for(section).present?
-  end
-
-  # FIXME: Rollify uses polymophism which doesn't work properly
-  # with STI on Section. Suggest using a collaborator model on sections instead
-  def collaborator_on?(section)
-    object.roles.exists?(resource_id: section)
-  end
-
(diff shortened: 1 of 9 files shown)


freshtonic:
Before: `Completed 200 OK in 1922ms (Views: 1761.1ms | ActiveRecord: 137.6ms)`

freshtonic:
Before: `Completed 200 OK in 1922ms (Views: 1761.1ms | ActiveRecord: 137.6ms)`

simonschwartz:
LGTM :)

app/decorators/user_decorator.rb
```diff
@@ -9,20 +9,8 @@ def display_name
...
-    pending_request_for(section).present?
-  end
-
-  # FIXME: Rollify uses polymophism which doesn't work properly
-  # with STI on Section. Suggest using a collaborator model on sections instead
-  def collaborator_on?(section)
-    object.roles.exists?(resource_id: section)
-  end
 
+  # TODO: THIS SHOULD NOT BE HERE
```
bjgordon: Then we should remove this, but lets worry about it later.

bjgordon:
LGTM

app/decorators/user_decorator.rb
freshtonic: I'll kill it now

PR #412, opened 2016-08-21
User feedback function SITES-624

coderdan:


app/controllers/feedback_controller.rb
```diff
@@ -0,0 +1,21 @@
+class FeedbackController < ApplicationController
+  def new
+    @feedback = Feedback.new(url: params[:url])
```
ltankey: Could we not instead use HTTP_REFERER?

db/schema.rb
```diff
@@ -22,12 +22,42 @@
...
     t.integer  "children_count"
-    t.datetime "created_at",                     null: false
-    t.datetime "updated_at",                     null: false
-    t.boolean  "placeholder",    default: false, null: false
+    t.datetime "created_at",     null: false
+    t.datetime "updated_at",     null: false
     t.index ["slug"], name: "index_categories_on_slug", using: :btree
   end
 
+  create_table "content_blocks", force: :cascade do |t|
```
ltankey: We need to remove this from schema.rb - it is an artefact from a production database dump and isn't actually a part of our application anymore.

When loading production data we need to take care to locally drop this table to avoid this.

app/controllers/feedback_controller.rb
coderdan: I'll have to check this to make sure it works with multiple form submissions. Should be ok. Good suggestion.

db/schema.rb
coderdan: Good catch. Its probably in my local DB. I'll drop the table.

ltankey:
LGTM

app/views/feedback/create.html.haml
```diff
@@ -0,0 +1,3 @@
+Thanks for your feedback
+
+= link_to "Return to gov.au", @feedback.url
```
tessereth: This could possibly be a redirect with a flash message rather than a whole page just to say "Thanks for your feedback." Just a thought though.

db/schema.rb
```diff
@@ -22,12 +22,34 @@
...
+    t.datetime "updated_at", null: false
+    t.index ["node_id"], name: "index_data_bundles_on_node_id", using: :btree
+  end
+
+  create_table "data_public_holidays", force: :cascade do |t|
+    t.string   "name"
+    t.date     "date"
+    t.datetime "created_at", null: false
+    t.datetime "updated_at", null: false
+  end
```
tessereth: The data_bundles and data_public_holidays tables are not created in this PR. Should they be here?

spec/controllers/feedback_controller_spec.rb
```diff
@@ -0,0 +1,63 @@
...
+        subject
+        expect(assigns(:feedback).url).to be_blank
+      end
+    end
+
+    context 'when a URL is passed' do
+      before  { request.env['HTTP_REFERER'] = '/foo' }
+      it      { is_expected.to be_success }
+
+      specify 'that no URL is set on the feedback instance' do
```
tessereth: I believe the `no` should be removed from this sentence

spec/models/feedback_spec.rb
```diff
@@ -0,0 +1,4 @@
+require 'rails_helper'
+
+RSpec.describe Feedback, type: :model do
+end
```
tessereth: If this file is empty, perhaps remove it?

app/views/feedback/create.html.haml
coderdan: I'm actually waiting on a bunch of content to go on this page. Once I have more detail I'll open another ticket and either add it or do as you suggest.

spec/fabricators/feedback_fabricator.rb
```diff
@@ -0,0 +1,2 @@
+Fabricator(:feedback) do
+end
```
tessereth: Feedback objects are expected to have an email and comment. They should probably be added here. That being said, you don't appear to be using this fabricator so you could also just remove it.

db/schema.rb
coderdan: Ah crap. They were from a spike and when I dumped the schema they ended up in there. I'll remove.

spec/controllers/feedback_controller_spec.rb
coderdan: Yep. English fail.

spec/models/feedback_spec.rb
coderdan: Was going to leave it as a placeholder but sure.

spec/controllers/feedback_controller_spec.rb
coderdan: Actually copy paste fail.

spec/fabricators/feedback_fabricator.rb
coderdan: Nope not using it.

app/views/feedback/new.html.haml
```diff
@@ -0,0 +1,6 @@
+%section.form_wrapper
+  = simple_form_for(@form, url: feedback_path) do |f|
+    = f.hidden_field :url
+    = f.input :email, as: :email
+    = f.input :comment, as: :text
+    = f.button :submit, 'Send Feedback', role: 'button'
```
tessereth: This page has the following issues:
* No h1
* No `class="content-main"` -> required for the proper button formatting to be applied
* No `#content` element -> required for "Skip to main content" link to work

Note: The `section` layout provides the last two but the `application` layout does not. Of course you can't use the section layout anyway but that may be why it's not doing what you expect.

You may wish to add the feedback page(s) to the accessibility tests by adding an `include_examples` to the bottom of `accessibility_spec.rb` to pick some of this up.

app/views/feedback/create.html.haml
tessereth: In that case, this page has all the same accessibility problems as the `new` page. Also, you probably meant the `Thanks...` and `link_to` to be in separate `<p>` tags so they end up on different lines.

tessereth: Now that I think about it some more, isn't it normal to redirect after a post? So the user doesn't resubmit the form if they refresh? So maybe this should be the show view rather than the create view.

app/views/feedback/new.html.haml
tessereth: Oh, and on the topic of `accessibility_spec.rb`, if you add the feedback pages to `config/sitemap.rb` (in the public group because they are) then the accessibility test that runs through a sitemap will also check them.

app/views/application/_feedback_link.html.haml
```diff
@@ -0,0 +1,2 @@
+.feedback{tabindex:1}
```
tessereth: I know you didn't introduce the `tabindex: 1` in this PR, but it should be removed and maybe you can do that here. It forces the feedback button to _always_ come first in the tab order and fails one of axe-core's best-practice tests.

app/views/feedback/new.html.haml
tessereth: Thanks for fixing the class and id issues but this page still needs an h1. As does the create page.

app/views/application/_feedback_link.html.haml
coderdan: Ah I missed it. That explains some other weird behaviour now. I can delete another tabindex later in the code too.

app/views/feedback/new.html.haml
coderdan: I know. I'm still waiting on content from the content people. I don't know what the H1 should say yet! :( I'll follow up.

tessereth:
Update changelog post 1.1.0 release

CHANGELOG.md
```diff
@@ -49,6 +50,7 @@ This file is influenced by http://keepachangelog.com/.
 ## [v1.0.2] - 2016-08-03
 ### Added
 - Transition links displayed on root page
+- Basic feedback functionality and admin
```
maxious: Duplicate of above.

app/controllers/feedback_controller.rb
```diff
@@ -0,0 +1,21 @@
+class FeedbackController < ApplicationController
+  def new
+    @feedback = Feedback.new(url: request.referrer)
+    @form = FeedbackForm.new(@feedback)
+  end
+
+  def create
+    @form = FeedbackForm.new(Feedback.new)
+    if @form.validate(feedback_params)
+      @form.save
```
tessereth: What if the form fails to save?

app/forms/feedback_form.rb
```diff
@@ -0,0 +1,9 @@
+class FeedbackForm < Reform::Form
+  property :url
+  property :email
+  property :comment
+  property :role
+  property :organisation
```
tessereth: Role and organisation are here but not in the administrate forms. Should they be?

app/views/feedback/new.html.haml
```diff
@@ -0,0 +1,15 @@
+
+%section#content.content-main.form_wrapper{role: "main"}
+  %h1 Your Feedback
```
tessereth: The capitalisation style used on the rest of the site, is capital first word and subsequent words in lowercase (unless their proper nouns). We should probably do that for all the headings/labels in this form as well.

app/views/feedback/new.html.haml
```diff
@@ -0,0 +1,15 @@
...
+  = simple_form_for(@form, url: feedback_path) do |f|
+    = f.input :comment, as: :text
+    = f.hidden_field :url
+    %h3 About You
```
tessereth: h2?

app/views/feedback/new.html.haml
coderdan: Well I copied it straight from the copy provided by the content design folk but sure I'll do that and let them know.

app/views/feedback/new.html.haml
coderdan: The h2 styling was bigger than the h1!

tessereth: Maybe it's not a heading at all. Is it a fieldset with a legend? Like in `nodes/edit.html.haml`?

coderdan: Good suggestion. I'll try it.

app/forms/feedback_form.rb
coderdan: Yar. Man I'm sucking at this.

app/controllers/feedback_controller.rb
coderdan: It won't because its valid. This is programming by contract and the recommended pattern for reform.

tessereth: Ah. Of course. Makes sense.

app/dashboards/feedback_dashboard.rb
```diff
@@ -0,0 +1,52 @@
...
+    :created_at
+  ].freeze
+
+  # SHOW_PAGE_ATTRIBUTES
+  # an array of attributes that will be displayed on the model's show page.
+  SHOW_PAGE_ATTRIBUTES = [
+    :id,
+    :email,
+    :role,
+    :organisation,
```
tessereth: You added role and organisation here but they should also be added in `ATTRIBUTE_TYPES`.

app/models/feedback.rb
```diff
@@ -0,0 +1,2 @@
+class Feedback < ApplicationRecord
```
micapam: I wonder if there could be some security implications here...

I was wondering if this might be useful for other govt sites / projects, e.g. for transitions. Then I realised that it may be possible to arrive at the feedback endpoint from _any_ site (unless I'm misreading this PR). The fact that we have a link 'return to gov.au' suggests that we might in certain circumstances end up with people thinking they are returning to gov.au if they arrive from a different referrer ... should we have a validation whitelist / regex on the URL, perhaps?

db/schema.rb
```diff
@@ -40,6 +50,17 @@
...
+    t.string   "code"
+    t.string   "accepted_token"
+    t.datetime "accepted_at"
+    t.string   "email"
+    t.datetime "created_at",     null: false
+    t.datetime "updated_at",     null: false
+    t.index ["accepted_token"], name: "index_invites_on_accepted_token", using: :btree
+    t.index ["code"], name: "index_invites_on_code", using: :btree
+  end
+
```
tessereth: Are these appearing from a different PR?

app/models/feedback.rb
coderdan: @micapam its a good point but I don't think we need to consider it in this PR. I would like to extract the feedback tool into its own service at which point we could consider these sorts of issues.

db/schema.rb
coderdan: Yes :( Too many parallel things!

tessereth:
LGTM

coderdan:
Bump LGTM

PR #416, opened 2016-08-22
[Sites-608] homepage updates

simonschwartz:
This PR adds author generated categories to the homepage along with some visual updates.

home-update-1


maxious:
LGTM

app/controllers/nodes_controller.rb
```diff
@@ -25,7 +25,9 @@ def home
     @ministers = Minister.all
     @departments = Department.all
     @agencies = Agency.all
-    @categories = STATIC_DATA[:categories] #TODO Replace once categories are proper data
+    @categories = Category.roots.where(:placeholder => false)
+    @categories_coming_soon = Category.roots.where(:placeholder => true)
+    # @categories_coming_soon = STATIC_DATA[:categories] #TODO Replace once categories are proper data
```
tessereth: Could possibly remove this comment?

app/views/templates/root_node.html.haml
```diff
@@ -2,17 +2,45 @@
 
 %article.content-listing
   %section.home-category-list
-    %h2 Information and services
+    %h2 What is GOV.AU Beta?
+    .abstract
+      %p
+        GOV.AU Beta is a small imagining of what a whole-of-government platform could look like. Its goal is to make Australian Government information and services simpler, clearer and faster.
```
tessereth: It would be nicer to word wrap these lines to 100 characters. Just for readability.

app/views/templates/root_node.html.haml
```diff
@@ -2,17 +2,45 @@
...
+        We're designing GOV.AU in phases and this Beta version is a very early release. This means we can launch new features swiftly, speak to users regularly and get feedback as we go – helping us create the best possible experience for everyone.
+
+    %h2 What’s available now?
+    .abstract
+      %p
+        GOV.AU Beta includes some examples of user-focussed information and services content that we’ve grouped into categories.
+      %p
+        We’ve also worked on some
+        %a{href: '/departments'}
+          agency ‘corporate’ content
```
tessereth: Could use `link_to` rather than an explicit `<a>`

config/static_data.yml
```diff
@@ -23,10 +23,6 @@ categories:
   - name: Immigration, visas and citizenship
     summary: >
       Australian citizenship, customs, visas, migration and tourism.
-  - name: Infrastructure and telecommunications
-    summary: >
-      Transport safety, television reception, telecommunications.
-    path: /categories/infrastructure-and-telecommunications
```
tessereth: Do we still use any of the categories in this file or are they now coming from the database? If not, maybe we should delete them all and update `config/initializers/static_data.rb` to not load them in?

spec/features/home_page_spec.rb
```diff
@@ -11,7 +11,7 @@
   end
 
   describe 'categories' do
-    it { expect(page).to have_link('Infrastructure and telecommunications',
-      href: '/categories/infrastructure-and-telecommunications') }
+    # it { expect(page).to have_link('Infrastructure and telecommunications',
+       # href: '/categories/infrastructure-and-telecommunications') }
```
tessereth: This test should be replaced with something appropriate to the new way categories work. Maybe create a placeholder and a non-placeholder category and check they turn up in the right places.

tessereth:
LGTM

PR #424, opened 2016-08-24
Remove rails_12factor gem

bjgordon:
Not needed in rails 5: https://github.com/heroku/rails_12factor#rails-5-and-beyond

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -27,8 +27,6 @@ gem 'dry-types'
 gem 'draper', git: 'https://github.com/coderdan/draper.git'
 gem 'devise'
 gem 'cf-app-utils' # cloudfoundry utils
-gem 'rails_serve_static_assets' # http://docs.cloudfoundry.org/buildpacks/ruby/ruby-tips.html#rails-4
-gem 'rails_12factor'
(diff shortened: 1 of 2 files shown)


maxious:
Do we also need to add the logging lines? https://github.com/heroku/rails_12factor#rails-5-and-beyond

(I don't think we're using Apache/NGINX)

bjgordon:
@maxious Those lines were already in `production.rb`. However you made me look closer, and we were also including `rails_serve_static_assets` gem which is not needed, I've removed.

maxious:
I think we *do* need rails_serve_static_assets because we're not using NGINX/Apache. Perhaps could test cf push somewhere with the gem removed?

bjgordon:
The README is a little unclear, but I'd thought that's what `config.public_file_server.enabled = ENV['RAILS_SERVE_STATIC_FILES'].present?` in `production.rb` did. We'd need to define that env var though. I'll test it.

bjgordon:
Worked for me with staging -green. As @maxious mentioned offline, the build-pack will set that env var.

tessereth:
LGTM

maxious:
LGTM

PR #425, opened 2016-08-24
Add closing </nav> tag

simonschwartz:
This PR fixes the broken layout on the Norther Territory daylight savings page caused by an open <nav> tag.


tessereth:
And this is why I like haml...

LGTM

bjgordon:
LGTM

Issue #415, opened 2016-08-22
Ability for author to create 'greyed out' hyperlinks with markdown


Alex-DTO:
As a content author
I want to be able to easily represent a dead/not yet valid hyperlink in body text to a reader
So that the reader is able to understand that there will be a link placed here at a later date

Content body text (along with other areas of the site) needs to be able to represent dead links to a user in order to show that there are to be links here at some point - this helps to expand the visibility of the content model, whereby a dead link infers that there has already been thinking about a page to sit behind that link.

Guides team have already released ability to create the dead link visuals using html in the markdown editor
http://gov-au-ui-kit.apps.staging.digital.gov.au/section-link-styles.html#guide-link-styles-3-placeholder-links

Requirement is to create markdown that will allow a content author to create this functionality without having to paste HTML code into the page for each link.

Alex-DTO:
Point for Discussion...

Currently in the editor we have used the following to denote a dead link:

[some text](#)

Suggestion is that ](#) is recognised to specify a dead link... This doesn't appear anywhere else in the content and would also save a lot of content debt in changing it to a different markdown instruction!

streetdaddy:
Triaged as SITES-627

bjgordon:
This was released in v.1.2.0

PR #428, opened 2016-08-24
SITES-627 Render placeholder links as placeholder-link spans

bjgordon:
Change placeholder links to [placeholder-link spans](http://gov-au-ui-kit.apps.staging.digital.gov.au/section-link-styles.html#guide-link-styles-3-placeholder-links) as requested in #415.
Have only implemented for the public markdown renderer, so it won't happen in the editor client-side preview. Am tempted to change to a regex approach, but it was simpler this way re-using code from the spike. And I didn't want [to add another problem](http://regex.info/blog/2006-09-15/247). Thoughts?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ This file is influenced by http://keepachangelog.com/.
 - Verification of identity when accessing protected pages
 - Phone number verification when changing numbers
 - Simple feedback system and admin
+- Rendering # links as placeholder spans
 
 ### Changed
 - Upgrade Rails to v5.0.0.1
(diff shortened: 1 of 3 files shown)


maxious:
I suspect a regex would be faster especially if this has to be called at every render time (premature optimization). 
And the preview could be done in JS. Beware cthulhu https://blog.codinghorror.com/parsing-html-the-cthulhu-way/

bjgordon:
Thanks, yep I'm going to avoid `problems++` and cthulhu by overriding the link render.

bjgordon:
Removed WIP - ready for review again.

tessereth:
LGTM

ltankey:
LGTM

PR #429, opened 2016-08-24
SITES-629 Use activerecord session store

maxious:
DB backed session store can invalidate logged out sessions server side. This also means the local cookie can be limited to the current browser session.

Unfortunately, this means every user of the website is getting a cookie and a session in the DB because of CSRF prevention. Not sure that's great for caching or performance.
There is a rake task to clean the session table db:sessions:trim but for now, there's a hack to do it each time an admin logs in.


tessereth:
Would something like this https://github.com/javan/whenever work for running the session trim task?

maxious:
Used https://github.com/jmettraux/rufus-scheduler to have variability in run time

ltankey:
LGTM

tessereth:
Would it make sense to have a test that logged out users are actually logged out?

maxious:
I don't think it can be tested other than by hand using cookie/db deletions.

tessereth:
Cool. Then LGTM.

PR #430, opened 2016-08-24
SITES-495 Navigation items at all levels are not using Design Guide styles to highlight the current page

streetdaddy:
Updated to match the ui-kit spec class names for current nav item:

http://gov-au-ui-kit-staging.apps.staging.digital.gov.au/section-navigation.html#guide-navigation-2-local-navigation

nb. I made this change content blind because I got stuck trying to log in to localhost and create the required page hierarchy 😞


ltankey:
@streetdaddy is this WIP?

streetdaddy:
@tessereth updated the nav template logic.

tessereth:
We also need to apply the `is-active` class to all parents of the current node. Would you be okay with me making my own PR or would you rather I explain how to find the parents of a node so you can do it?

streetdaddy:
Outdone by #446

PR #434, opened 2016-08-25
Feature/beta invites

coderdan:
WIP


app/controllers/application_controller.rb
```diff
@@ -1,6 +1,22 @@
 require 'digest/sha1'
 
 class ApplicationController < ActionController::Base
+  before_action :beta_access
+
+  def beta_access
+    redirect_to go_invites_path unless has_beta_access?
```
maxious: What about letting people log in?

app/views/invites/show.html.haml
```diff
@@ -0,0 +1,6 @@
+
```
maxious: SHould this be in editorial? Who has access?

spec/models/invite_spec.rb
```diff
@@ -0,0 +1,5 @@
+require 'rails_helper'
```
maxious: Please make specs for creating, sending and using/failing to access via invites

bjgordon:
Closing as this feature is parked for now

PR #435, opened 2016-08-25
SITES-590 Bypass complete_two_factor_setup in dev

bjgordon:
A quick fix for
image


app/controllers/application_controller.rb
```diff
@@ -30,17 +30,22 @@ def decorated_current_user
...
-      destroy_user_session_path
-    ]
-
-    if current_user && !current_user.account_verified && current_user.confirmed_at
-      # Users need to be able to verify their tfa code
-      unless request.path.start_with?(users_two_factor_setup_path) ||
-        whitelist.include?(request.path)
-
-        redirect_to new_users_two_factor_setup_path
+  if Rails.env.development?
```
tessereth: Should probably also check for `Rails.configuration.force_2fa.present?`. If it's present, you should do the 2fa even in dev.

bjgordon:
Thanks @tessereth, added that in

tessereth:
LGTM

maxious:
LGTM

config/initializers/two_factor_auth.rb
```diff
@@ -0,0 +1,4 @@
+
+Rails.application.configure do
+  config.use_2fa = !Rails.env.development? || ENV['FORCE_2FA'].present?
```
ltankey: Maybe we can use something like

`config.use_2fa = (Rails.env.development? && ENV['FORCE_2FA'].present?) || !ENV['DISABLE_2FA']`

.env.sample
```diff
@@ -18,3 +18,4 @@ CONTENT_ANALYSIS_BASE_URL=http://localhost:3001
 
 # Leave commented out if not wanted
 # FORCE_2FA=true
+# DISABLE_2FA=true
```
ltankey: We should probably add a comment heading for this envvar that it is a prod-only environment variable that isn't required to be set as part of a developer's `.env` file. Ideally we should capture this in a readme, but in absence of a section for that, this would be good enough.

PR #436, opened 2016-08-25
Remove check on CONTENT_ANALYSIS_BASE_URL

bjgordon:


tessereth:
LGTM

ltankey:
Perhaps an opportunity to remove the helper outright. I suspect when we revisit this, we'd approach it differently anyway.

bjgordon:
Ok will do

bjgordon:
This has lead to complete removal of CAS. Will open a new PR and branch...

PR #441, opened 2016-08-26
[WIP] SITES-632 Create RootSection to allow authors to create root level pages

maxious:
RootNode child nodes do appear as root level pages but without a section, you can't find them in editorial or give access to edit those pages etc. 
This PR attempts to add a special section type that can hold a RootNode and relax some of the UI/validations to let users easily manage root level pages.

### DIFF
--- a/app/controllers/admin/root_sections_controller.rb
+++ b/app/controllers/admin/root_sections_controller.rb
@@ -1,0 +1,23 @@
+module Admin
+class TopicsController < Admin::ApplicationController
+    # To customize the behavior of this controller,
+    # simply overwrite any of the RESTful actions. For example:
+    #
+    # def index
+    #   super
+    #   @resources = Topic.all.paginate(10, params[:page])
+    # end
+
+    # Define a custom finder by overriding the `find_resource` method:
+    # def find_resource(param)
+    #   Topic.find_by!(slug: param)
+    # end
+
+    # See https://administrate-docs.herokuapp.com/customizing_controller_actions
+    # for more information
+
+    def find_resource(param)
+      RootSection.find param
+    end
+  end
+end
--- a/app/models/general_content.rb
+++ b/app/models/general_content.rb
@@ -3,7 +3,11 @@ class GeneralContent < Node
 
   validates_with SectionHeritageValidator
   validates_presence_of :section
-  validates :parent, ancestry_depth: { minimum: 2 }
+  validates :parent, ancestry_depth: { minimum: 2 }, if: :not_root_page?
+
+  def not_root_page?
+    not section.is_a? RootSection
+  end
 
   def available_options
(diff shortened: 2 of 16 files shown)


app/controllers/admin/root_sections_controller.rb
```diff
@@ -0,0 +1,23 @@
+module Admin
+class TopicsController < Admin::ApplicationController
```
freshtonic: The class name does not match the file name.

app/models/general_content.rb
```diff
@@ -3,7 +3,11 @@ class GeneralContent < Node
 
   validates_with SectionHeritageValidator
   validates_presence_of :section
-  validates :parent, ancestry_depth: { minimum: 2 }
+  validates :parent, ancestry_depth: { minimum: 2 }, if: :not_root_page?
+
+  def not_root_page?
+    not section.is_a? RootSection
```
freshtonic: use `!` not `not`. Keyword versions of the ruby operators have different precedence rules and a generally frowned upon.

freshtonic:
@maxious this is a clever approach to having non-section nodes at the root level but it builds upon (and increases) the "sections are special" tech debt. Can we not achieve the same thing by allowing GeneralContent nodes to be directly attached to the root, and update the UI to be aware of them?

maxious:
GeneralContent nodes can be directly attached to the root but this was a neat way to get a UI. That being said, I'm starting to doubt that having more root level pages is the correct approach for this task.

freshtonic:
The app implementation has a strong 1-1 correspondence to the IA, and I think this has bitten us a bit. This is why Sections were first-class domain entities instead of just Nodes. It's also why there are weird UI quirks in the editorial (doubled up on breadcrumbs for SectionHome, and "am I on the section or section home node" quirk).

I'd like to be able to defer this story until _after_ we have refactored the Section/SectionHome dichotomy. Maybe we can get some time tomorrow to discuss what needs to be done and estimate. It's long overdue IMHO.  @coderdan what do you think?

maadonna:
Can you loop me in when you do talk about it (though if it's tomorrow I might be out with Thinkplace doing research). I'd like to understand why there might need to be pages (without being in a section) at the root. I'm not sure whether there is a need, or whether authors are asking for pages without thinking through whether they better belong in a section. I'm fairly confident we don't want distributed authors just creating pages at the top level.

freshtonic:
@maadonna it's it allow the `/about-us` page to be edited. Currently it's hard coded. I expect that creating root level pages would require special/admin privileges.

maadonna:
No worries. As I noted on the jira issue that it's being done in, I think all the footer pages should be in a 'About GOV.AU' section anyway, and then 'about gov.au' is the landing page for the section. But that's probably longer term than the immediate hack

PR #442, opened 2016-08-26
[SITES-648] Bugfix/sms send via jobs

ltankey:
Moves SMS dispatching to Rails ActiveJob instead of a Ruby thread.


maxious:
LGTM

spec/jobs/send_two_factor_authentication_code_for_job_spec.rb
```diff
@@ -0,0 +1,60 @@
...
+      stub_request(:post, Rails.configuration.sms_send_message_url).
+          with(:body => "{\"to\":\"#{complete_user.unconfirmed_phone_number}\",\"body\":\"Your GOV.AU two-factor authentication code is #{complete_user.unconfirmed_phone_number_otp}\"}",
+               :headers => {'Authorization'=>'Bearer somereallyspecialtoken', 'Content-Type'=>'application/json'}).
+          to_return(:status => 200, :body => "", :headers => {}
+          )
+    }
+
+    it 'should authenticate the sms service and send the sms' do
+      subject
+      expect(send_sms_request).to have_been_requested
```
bjgordon: Missing a `expect(valid_authenticate_request).to have_been_requested`? Could also DRY this spec up...

ltankey: Actually, and probably more confusingly, the class variable for the test persists which means that the authentication only needs to occur once. Asserting the authenticate call you mention causes the test to fail, because it doesn't get called (the class still has a valid token).

Perhaps we should have some spin-up/tear-down methods to reset the class variables?

bjgordon:
LGTM

PR #445, opened 2016-08-29
[SITES-647] Allow creating users through editorial

tessereth:
Creating a new user sends an invite email to the user. When they click the link it sends them to the confirmation page which redirects to the password reset page.

This PR does not remove the registration pages. That's coming later.

image

### DIFF
--- a/app/controllers/editorial/users_controller.rb
+++ b/app/controllers/editorial/users_controller.rb
@@ -1,0 +1,35 @@
+module Editorial
+  class UsersController < EditorialController
+
+    before_action ->() { authorize! :create, :user }, only: [:new, :create]
+
+    def new
+      @form = UserForm.new(User.new)
+    end
+
+    def create
+      @form = UserForm.new(User.new)
+      if @form.validate(user_params)
+        @form.sync
+        tmp_password = SecureRandom.base64(12)
+        user = @form.model
+        user.password = tmp_password
+        user.password_confirmation = tmp_password
+        user.skip_confirmation_notification!
+        if user.save
+          user.send_invite_email
+          flash[:notice] = 'An email has been sent to the user inviting them to join GOV.AU'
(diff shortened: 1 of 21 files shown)


maxious:
LGTM

app/controllers/editorial/users_controller.rb
```diff
@@ -0,0 +1,35 @@
...
+    end
+
+    def create
+      @form = UserForm.new(User.new)
+      if @form.validate(user_params)
+        @form.sync
+        tmp_password = SecureRandom.base64(12)
+        user = @form.model
+        user.password = tmp_password
+        user.password_confirmation = tmp_password
```
ltankey: Can a nil password be used instead?

app/views/devise/passwords/edit.html.haml
```diff
@@ -8,5 +8,5 @@
       = f.input :password, label: "New password", required: true, autofocus: true, hint: ("#{@minimum_password_length} characters minimum" if @minimum_password_length)
       = f.input :password_confirmation, label: "Confirm your new password", required: true
     .form-actions
-      = f.button :submit, "Change my password"
+      = f.button :button, "Change my password"
```
ltankey: Does this set the role? Or the class?

app/controllers/editorial/users_controller.rb
tessereth: It fails model validations without a password

app/views/devise/passwords/edit.html.haml
tessereth: It makes a `<button>` rather than a `<submit>`. You need either that or `role="button"` for the ui-kit styles to apply properly and I have a vague preference for `<button>`.

ltankey:
Got it, LGTM

PR #447, opened 2016-08-29
[SITES-638] Import database from S3

bjgordon:


lib/tasks/db.rake
```diff
@@ -1,49 +1,84 @@
...
-  desc 'Scrubs database dump file and uploads to S3'
-  task :s3upload => :dump do
-    begin
-      scrubbed_file = File.join(Dir.tmpdir, 'db_scrubbed.sql')
-      s3_key = "db_#{ENV['APP_DOMAIN']}.sql"
-      s3_bucket = ENV['AWS_DB_S3UPLOAD_BUCKET']
+  desc 'Import a database dump from S3'
+  task :'s3:import', [:key] => :environment do |t, args|
+    unless args.key.present?
+      abort 'You must specify the S3 key to import e.g. rake db:download[db_gov-au-beta.example.gov.au.sql]'
```
tessereth: Should this be `s3:import` not `db:download`? Or `db:s3:import`?

lib/tasks/db.rake
```diff
@@ -1,49 +1,84 @@
...
+    end
+
+    Rake::Task['db:drop'].invoke
+    Rake::Task['db:create'].invoke
```
tessereth: It might be a good idea to drop after downloading the database incase the download fails.

lib/tasks/db.rake
bjgordon: Oops I'll change the message to `db:s3:import`

tessereth:
LGTM

freshtonic:
LGTM

PR #448, opened 2016-08-29
[SITES-653] Confirm users when they are created

bjgordon:
This prevents confirmation emails being sent when creating the users in seeds.


tessereth:
LGTM

freshtonic:
LGTM

PR #449, opened 2016-08-29
SITES-386 allow section home node reordering

freshtonic:
Reordering was prevented because of parent ID validations on the
SectionHome, so:

1) Node base class has a `reparentable?` method which returns true,
   which is overridden in SectionHome to return false.
2) Parent drop down on Node metadata form is only rendered when node is
   reparentable.
3) SectionHome has a new validation that prevents the parent_id from
   being changed.  This prevents the parent from being changed by
   hacking the form.

With the correct parent validations, the form can be saved and thus
reordering works.


tessereth:
Looks good but it would be nice to have a test that you can reorder the children of a section home node now. Also changelog.

freshtonic:
@tessereth changelog updated. Regarding a test that nodes can be reordered on a section home: I've specced the validations for SectionHome & written a feature spec for the metadata form. I know why it was failing to reorder and the tests I wrote cover that.

tessereth:
Okay. LGTM.

maxious:
LGTM

PR #450, opened 2016-08-29
Scrub out session table contents

maxious:
Aside from potentially sensitive contents, they're unnecessary when moving between systems.


ltankey:
LGTM

bjgordon:
LGTM

PR #451, opened 2016-08-29
[SITES-616] Remove sign-up routes and links

tessereth:


spec/routing/routes_spec.rb
```diff
@@ -0,0 +1,11 @@
+require 'rails_helper'
+
+describe 'routes', type: :routing do
```
freshtonic: I think you can kill this test. It was useful for you in development but will never be useful again.

The thing with testing for things that _should not_ happen is that it's an infinite set. The app shouldn't Rick Roll people every third thursday but there's no test for that either 😋

tessereth: The argument for adding it is that devise might decide to bring it back for whatever reason. So because we're dependent on a third party gem, we want to be extra sure. That being said I'm not particularly enamoured to it.

tessereth: Removed.

freshtonic: Inclusion of Devise modules (and thus their routes) has always been explicit so I think we're safe.

LGTM.

freshtonic:
(╯°□°)╯︵ W⊥⅁⅂

LGTM﻿ ノ( ゜-゜ノ)

.gitignore
```diff
@@ -43,6 +43,7 @@ bower.json
 # IntelliJ IDEA files
 .idea
 **.iml
+railways.cache
```
maxious: 🚂 ?

maxious:
:wave: LGTM

.gitignore
tessereth: railways is an intellij plugin to make dealing with routes easier. These are all links that take you to the controller method.

screen shot 2016-08-30 at 9 08 43 am

PR #454, opened 2016-08-30
Ensure all content is within an aria landmark

tessereth:
Note: The changelog already has an entry covering accessibility


ltankey:
LGTM

maxious:
LGTM

PR #456, opened 2016-08-30
[SITES-641] Feature/sms or authenticator 2fa

ltankey:
This introduces a 2FA setup wizard for users who have no yet verified their account (on account setup completion). The SMS flow is the same as what already exists, however it is now tied into a wizard-style process with an additional step to configure an authenticator.

image

image


bjgordon:
To make it easy for people testing two-factor locally, please add `TFA_SECRET_KEY` to .env.sample

app/controllers/users/two_factor_setup/authenticator_setup_controller.rb
```diff
@@ -0,0 +1,75 @@
...
+
+        return false
+      end
+
+
+      def generate_qrcode
+        @totp_secret = session[:setup_totp]
+
+        if @totp_secret
+          @qrcode_uri = "otpauth://totp/#{current_user.email}?secret=#{@totp_secret}&issuer=gov.au"
```
bjgordon: I think it'll be easier to handle staging/prod etc to change issuer to `ENV['APP_DOMAIN']`

app/controllers/users/two_factor_setup/authenticator_setup_controller.rb
```diff
@@ -0,0 +1,75 @@
...
+          @form_error = "Your code didn't work. Please try again."
+          render :new and return
+        end
+
+        redirect_to new_users_two_factor_setup_authenticator_path
+      end
+
+
+      def resend_code
+        session[:setup_totp] = current_user.generate_totp_secret
```
bjgordon: You could just clear `session[:setup_totp]` here, it will be (re)created in #new.

app/controllers/users/two_factor_setup/authenticator_setup_controller.rb
ltankey: Good catch!

app/controllers/users/two_factor_setup/authenticator_setup_controller.rb
ltankey: Good call.

app/controllers/users/two_factor_setup/authenticator_setup_controller.rb
```diff
@@ -0,0 +1,76 @@
...
+
+      def resend_code
+        session.delete :setup_totp
+        redirect_to new_users_two_factor_setup_authenticator_path
+      end
+
+
+      private
+      def authenticate_code(key, code)
+        unless code.nil? || key.nil?
```
bjgordon: https://github.com/airbnb/ruby#unless-with-multiple-conditions

ltankey: Good pick-up - changing.

app/controllers/users/two_factor_setup_controller.rb
```diff
@@ -1,65 +1,49 @@
...
     end
 
 
-    # Send SMS and redirect to confirm if valid number
-    def create
-      unless params[:phone_number].nil?
-        current_user.unconfirmed_phone_number =  params[:phone_number]
+    def choice
+      session[:setup_2fa] = []
+      redirect_to users_two_factor_setup_path and return if params[:validation].nil?
```
bjgordon: Perhaps also show an error/flash if they don't select either option. Or disable the button until one is checked?

app/controllers/users/two_factor_setup_controller.rb
```diff
@@ -1,65 +1,49 @@
...
+      redirect_to users_two_factor_setup_path
     end
 
 
     private
-    # Don't let users skip ahead of being sent an actual code
-    def verify_direct_otp_sent
-      if current_user.unconfirmed_phone_number_otp == nil
-        redirect_to new_users_two_factor_setup_path
+    def redirect_confirmed_users
```
bjgordon: Rename to `redirect_verified_users`?

app/views/users/two_factor_setup/authenticator_setup/new.html.haml
```diff
@@ -0,0 +1,40 @@
...
+  .callout
+    %h3= @totp_secret
+
+  %section.sign-in-form
+    .sign-in-footer-links
+      .background-border-heading
+        %span.inner
+          Or
+
+  .qrcode
```
bjgordon: I think the qrcode should be displayed first. Would anyone ever manually type the code in??

app/controllers/users/two_factor_setup_controller.rb
ltankey: Good idea, added some simple JS and error handling if form submission happens anyway.

app/views/users/two_factor_setup/authenticator_setup/new.html.haml
ltankey: Maybe - hard to say without some user research though. The code is the actual secret-key, so in this context it's the actual value being used to generate the TOTPs. 

But perhaps this is beside the point. I suppose the use-case is some folks _might_ write it down or store the key somewhere securely. Happy to shift it though if we think the QR code should be more prominent.

app/views/users/two_factor_setup/sms_setup/new.html.haml
```diff
@@ -9,7 +9,7 @@
     Keep your account secure by using both your password and mobile phone
```
bjgordon: I think this content from when sms was the only 2 factor should change now, it reads a bit funny next to the authenticator setup page. But could also wait for proper content advice.

app/views/users/two_factor_setup/authenticator_setup/new.html.haml
bjgordon: Personally I usually just screenshot the qr code if I want a backup :-) But good points, just leave it and let's wait for research, unless someone else has a stronger opinion.

app/controllers/users/two_factor_setup/authenticator_setup_controller.rb
```diff
@@ -0,0 +1,81 @@
+module Users
+  module TwoFactorSetup
+    class AuthenticatorSetupController < TwoFactorVerificationController
+      before_action :authenticate_user!
```
bjgordon: I _think_ you can put `before_action :authenticate_user!` into TwoFactorVerificationController, and remove it from the other two_factor_setup controllers?

maxious:
LGTM

app/controllers/users/two_factor_setup_controller.rb
```diff
@@ -1,65 +1,54 @@
...
 
 
-    # Give form to input one-time-password
-    def confirm
-    end
-
+    def continue_setup
+      unless session[:setup_2fa].blank?
+        # Redirect to setup of session[:setup_2fa].first
+        redirect_to send("new_users_two_factor_setup_#{session[:setup_2fa].shift}_path") and return
```
tessereth: Is there any chance the session parameter here could come from user input?

tessereth:
In the interests of expediency, LGTM

PR #459, opened 2016-08-31
Better request and model change logging

maxious:
Uses https://github.com/roidrage/lograge to allow multiple future key/value/nested output formats in the future.

Request logging is thus condensed allowing the addition of fields like referer/host etc.

Model events trigger log events, user model changes trigger additional events for email/password changes.


app/controllers/admin/application_controller.rb
```diff
@@ -10,6 +10,12 @@ class ApplicationController < Administrate::ApplicationController
 
     before_action ->() { authorize! :manage, :all }
     before_action :complete_two_factor_setup
+    before_filter :set_current
```
ltankey: I think `before_filter` is deprecated for `before_action`

app/controllers/users/sessions_controller.rb
```diff
@@ -0,0 +1,17 @@
+class SessionsController < Devise::SessionsController
+  after_filter :log_failed_login, :only => :new
```
ltankey: `after_creation` should be the non-deprecated method.

app/models/user.rb
```diff
@@ -14,6 +14,34 @@ class User < ApplicationRecord
...
+      event = 'user_password_update'
+    elsif changes['reset_password_token']
+      event = 'user_password_reset'
+    elsif changes['email']
+      event = 'user_email_update'
+    elsif changes
+      event = 'user_update'
+    end
+    if event
+      data = {event: 'user_update', record_id: self.id, record_type: self.class, record_name: self.email, changes: changes}
```
ltankey: Should this use `event` as the value to the event key?

maxious: Shoudl there be an event for user login?

app/controllers/sessions_controller.rb
```diff
@@ -0,0 +1,22 @@
+class SessionsController < Devise::SessionsController
+  after_filter :log_failed_login, :only => :new
```
ltankey: We should probably use the non-deprecated `after_action`

app/controllers/nodes_controller.rb
```diff
@@ -6,7 +6,9 @@ class NodesController < ApplicationController
   decorates_assigned :menu_nodes, with: NodeDecorator
 
   def show
+    Rollbar.silenced {
```
freshtonic: Use `do` ... `end` for multi-line blocks

app/models/user.rb
```diff
@@ -51,9 +79,9 @@ def collaborator_on?(section)
   # Required hook for two_factor_authentication gem
   def send_two_factor_authentication_code(code)
     SendTwoFactorAuthenticationCodeForJob.perform_later(
-       :phone_number.to_s,
-       :direct_otp.to_s,
-       self
+        :phone_number.to_s,
```
freshtonic: 2 character indentation please :)

app/views/layouts/root.html.haml
```diff
@@ -5,7 +5,7 @@
     .wrapper
 
       .govau--logo.home
-        = image_tag 'icons/logo-crest-2x.png', width: '65'
+        = image_tag 'icons/logo-crest-2x.png', width: '65', alt: "GOV.AU crest logo"
```
freshtonic: Does this change belong in this PR?

app/helpers/logging_helper.rb
```diff
@@ -0,0 +1,34 @@
+module LoggingHelper
```
freshtonic: The thread local data should be cleared after the HTTP request has been completed. This should be done in an `ensure` block so that no matter what errors are raised the data is cleared.

freshtonic: To be honest, I'm not even certain that this thread local approach is safe. Puma uses event machine (I think) and I'm not 100% certain that an HTTP request is always processed inside one single thread.

freshtonic: Doesn't look like Puma depends on EventMachine.

freshtonic: @maxious this gem might help for keeping thread locals tidy across requests: http://stackoverflow.com/a/13930563/4570488

app/controllers/admin/application_controller.rb
```diff
@@ -10,6 +10,12 @@ class ApplicationController < Administrate::ApplicationController
 
     before_action ->() { authorize! :manage, :all }
     before_action :complete_two_factor_setup
+    before_action :set_current
+
+    def set_current
+      LoggingHelper.current_user = current_user
+      LoggingHelper.current_request = request
+    end
```
freshtonic:
```

File: app/views/layouts/root.html.haml (hunk 7 above)
Comment (maxious) [inline app/views/layouts/root.html.haml]: Yep, found during testing so might as well fix ASAP

**Comment by maxious** [comment]
Current user/request now stored in safer way with https://github.com/ElMassimo/request_store_rails

File: app/controllers/application_controller.rb [hunk 10]
```diff
@@ -7,6 +7,12 @@ class ApplicationController < ActionController::Base
   protect_from_forgery with: :exception
   before_action :complete_two_factor_setup
   helper_method :decorated_current_user
+  before_action :set_current
+
+  def set_current
```
coderdan: Perhaps the same around_filter idea @freshtonic suggested earlier.

app/controllers/news_controller.rb
```diff
@@ -38,6 +38,8 @@ def show
 
   private
   def set_section
-    @section = SectionHome.find_by!(slug: params[:section]).section
+    Rollbar.silenced do
```
coderdan: Why do we not want rollbar logging here?

app/controllers/sessions_controller.rb
```diff
@@ -0,0 +1,22 @@
+class SessionsController < Devise::SessionsController
+  after_action :log_failed_login, :only => :new
+
+  def create
+    super
+    LoggingHelper.log_event(request,current_user,{event:"user_login"})
```
coderdan: There is a lot of repeated code here. Perhaps we could create a method in application controller like:

    def log_event(event, attrs)
      LoggingHelper.log_event(request, current_user, {event: event}.merge(attrs))
    end

Then you can just do:

    log_event('user_login')
    log_event("user_failed_login", user:request.filtered_parameters["user"])

app/controllers/news_controller.rb
maxious: We want the RecordNotFound exception to be handled by callers eg. Controllers so they can 404. But don't need rollbar spammed by 404 errors like /favicon.ico

app/controllers/sessions_controller.rb
maxious: 👍

freshtonic: @maxious with the `around_action` thing, just be aware of the controller hierarchy when you  implement it. If you do it in `ApplicationController` _and_ `EditorialController` you may end up doubling up on logging.

app/controllers/news_controller.rb
coderdan: Looking into the rollbar docs we might be better achieving this with an exception level filter (see https://rollbar.com/docs/notifier/rollbar-gem/#advanced-usage).

The issue I have with silencing rollbar is that if there is an exception raised other than a 404 we'll miss it.

app/helpers/logging_helper.rb
coderdan: Perhaps instead of using a module you should use a class. The way you're using it isn't very module like. Make it a class and put in lib would be my suggestion.

coderdan: Perhaps name it EventLog or something too.

app/models/concerns/audit_logging.rb
```diff
@@ -0,0 +1,28 @@
...
+    audit(self, "create_record")
+  end
+  def after_destroy
+    audit(self, "destroy_record")
+  end
+
+  def audit(record,action)
+    data = {event:action, record_id: record.id, record_type: record.class}
+    #get current user and request from the Thread
+    LoggingHelper.log_event(LoggingHelper.current_request, LoggingHelper.current_user, data)
```
coderdan: So AFAICT you're only including the AuditLogging module in controllers, right? If this is the case then you don't need to store the current_request or current_user on the thread to share it as they will already be in the controller.

You can do:

    def audit(record, action)
      data = {event:action, record_id: record.id, record_type: record.class}
      LoggingHelper.log_event(request, current_user, data)
    end

maxious: This Audit logging is a Model Concern. 
Models don't have those request/user variables hence the need to stash them somewhere in controllers for later use.

app/models/user.rb
```diff
@@ -14,6 +14,34 @@ class User < ApplicationRecord
   validates :phone_number, aus_phone_number: true
   validates :unconfirmed_phone_number, aus_phone_number: true
 
+  after_save :after_save
+
+  def after_save
```
coderdan: I'd recommend extracting all of this code out into a service object. It makes it easier to test and read and its reusable.

    def after_save
      RecordChangeLogger.new(self).perform
    end

    class RecordChangeLogger
      REDACT_KEYS = %w(encrypted_password ...)

      def initialize(record)
        @record = record
        @changes = determine_changes
      end

      def perform
        if @changes['encrypted_password']

        # etc
      end
 
      private
      def determine_changes
        self.changes.inject({}) do |h, (k, v)|
          if REDACT_KEYS.include?(k)
            h[k] =['REDACTED', 'REDACTED']
          else
            h[k] =v
          end
          h
        end
      end
    end

app/models/user.rb
```diff
@@ -14,6 +14,34 @@ class User < ApplicationRecord
...
+    elsif changes['reset_password_token']
+      event = 'user_password_reset'
+    elsif changes['unconfirmed_email']
+      event = 'user_email_update'
+    elsif changes
+      event = 'user_update'
+    end
+    if event
+      data = {event: event, record_id: self.id, record_type: self.class, record_name: self.email, changes: changes}
+      LoggingHelper.log_event(LoggingHelper.current_request, LoggingHelper.current_user, data)
```
coderdan: Really this violates MVC. Do we need to store the current request and current_user here? What if the change is being done on the console?

I think controller actions should be used to track that information.

config/initializers/filter_parameter_logging.rb
```diff
@@ -2,3 +2,7 @@
 
 # Configure sensitive parameters which will be filtered from the log file.
 Rails.application.config.filter_parameters += [:password]
+Rails.application.config.filter_parameters += [:password_confirmation]
```
coderdan: You could probably do all of these on one line.

spec/helpers/logging_helper_spec.rb
```diff
@@ -0,0 +1,13 @@
+require 'rails_helper'
+
+RSpec.describe LoggingHelper do
+
+  describe '#logging_messages' do
+    context 'logs a message'
```
coderdan: Are you missing a do here?

spec/helpers/logging_helper_spec.rb
```diff
@@ -0,0 +1,13 @@
...
+    it {
+      expect(Rails.logger).to receive(:info).with(/event=test_event/)
+      LoggingHelper.log_event(LoggingHelper.current_request, LoggingHelper.current_user, {event: "test_event"})
```
coderdan: Will current_request and current_user be nil here? Perhaps you should test with real values?

app/helpers/logging_helper.rb
maxious: A class in lib would need to be explicitly required in each file?

app/controllers/news_controller.rb
maxious: Having looked at the silenced block, I think it can be altered to only handle specific exceptions https://github.com/rollbar/rollbar-gem/blob/ffb006fa7bb900b82f76af7a3d127254dd9add25/lib/rollbar.rb#L89

app/models/user.rb
maxious: Okay, latest version has this abstraction.

ltankey:
LGTM

tessereth:
LGTM

app/services/user_change_logger.rb
```diff
@@ -0,0 +1,36 @@
+class UserChangeLogger
```
coderdan: This class should really have a spec

maxious: The behaviour is tested as part of user_spec.rb including the redaction which is a private method.

app/controllers/admin/application_controller.rb
```diff
@@ -10,6 +10,17 @@ class ApplicationController < Administrate::ApplicationController
 
     before_action ->() { authorize! :manage, :all }
     before_action :complete_two_factor_setup
+    around_action :setup_logging
+
+    def setup_logging
```
freshtonic: should be a private method or anonymous

maxious: I tried an anonymous method and it told me it could not yield.

freshtonic: Maybe a block can't take a block as an argument? Not worth worrying about for now.

PR #460, opened 2016-08-31
[SITES-664] remove 2fa grace window and force 2fa on every login

ltankey:
This PR will ensure 2FA is required for _every_ login, and that there is no grace window.


tessereth:
LGTM

bjgordon:
LGTM

PR #467, opened 2016-09-01
SITES-671 Add session admin dashboard

bjgordon:
Im not showing the session id or data, so this really only gives us an approximate count of session rows (using the row and page count) 

image


ltankey:
Changelog? I know it's not user facing, but it might be useful at some point.

ltankey:
LGTM with comments aside

maxious:
LGTM

Issue #462, opened 2016-09-01
Single quotes in markdown links render as #39;


nicholasgkdto:
When you add a ' in to the text for a link it shows up as `#39;`

ie 
`the cat's foot and [the dog's leg](http://www.example.com)`

show up as 

the cat's foot and [the dog#39;s leg](http://www.example.com)

only happens in links

see example 
platform/attorney-general-s-department/contact-us#social-media

ltankey:
Confirmed, -> SITES-674.

bjgordon:
Fixed in v1.3.0

Issue #461, opened 2016-09-01
Can not place links on images in markdown.


nicholasgkdto:
When I try to add a link to an image with markdown, it renders the html code and not the image.

Markdown i use:
`[![My alt text](https://avatars2.githubusercontent.com/u/19787722?v=3&s=88)](http://www.google.com)`

Expected results
[![My alt text](https://avatars2.githubusercontent.com/u/19787722?v=3&s=88)](http://www.google.com)

Results I get
[`<img src="https://avatars2.githubusercontent.com/u/19787722?v=3&s=88" alt="My alt text">`](http://www.google.com)

an example can be see here 
platform/department-of-communications-and-the-arts/news/2016-07-28-panel-presentation-the-art-of-change

I had to change it to just a link, but if you look at some of the page in the history it will be there

ltankey:
Confirmed - looks like a bug with our rendered output. Potentially related to our usage of RedCarpet - doesn't seem to happen in a kramdown renderer. SITES-672

bjgordon:
Fixed in v1.3.0

PR #471, opened 2016-09-02
SITES-672 SITES-674 Disable attribute value escaping in markdown link renderer

bjgordon:
This addresses #461 and #462.


ltankey:
Can we adjust the changelog entry just so that users who have reported this know that it's fixed? 
`Markdown links support images as a link type` or something.

Otherwise LGTM

tessereth:
LGTM

PR #472, opened 2016-09-05
SITES-585 Soft-delete users

freshtonic:


ltankey:
LGTM

maxious:
LGTM

spec/controllers/application_controller_spec.rb
```diff
@@ -36,7 +36,7 @@ def show
     describe "when a new version of the application is deployed" do
 
       it "busts the ETAG" do
-        controller.stub(:etag_seed).and_return "CURRENT_SHA1"
+        allow(controller).to receive(:etag_seed).and_return("CURRENT_SHA1")
```
bjgordon: I assume you changed this to get rid of the deprecation warning, if so there are other uses of `stub` in this file you could change now as well.

freshtonic: Hmm weird. I was only getting a single deprecation warning on that line so I didn't notice there were other usages.

PR #477, opened 2016-09-05
SITES-656 redirect authentication failures to /users/sign_in

freshtonic:
Authentication behaviour is now handled by using the AccessDeniedConcern
in appropriate controllers. This centralises the behaviour for
consitency. The behaviour is:

- When a user is not authenticated:

  If they try to access a page that requires authorisation they will be
  redirected to /users/sign_in

  If it's a public page the request will succeed

- When a user is authenticated:

  If the user is also authorised the request will succeed.
  If they are not authorised the request will fail with a 401.

The flash messages are also set by the helper.

I deleted some of the specs for various controllers that were testing
the very rudimentary behaviour of how authentication and authorisation
failures were handled.  The basics are now tested in the spec for the
AccessDeniedConcern itself so I removed the duplication.

### DIFF
--- a/app/views/errors/unauthorized.html.haml
+++ b/app/views/errors/unauthorized.html.haml
@@ -1,0 +1,7 @@
+= content_for :title do
+  You are not authorised to access this page (401)
+
+%article.content-main.callout--calendar-event
+  %h1 You are not authorised to access this page (401)
+  %p If you are the application owner check the logs for more information.
+
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ This file is influenced by http://keepachangelog.com/.
 ### Changed
 
 ### Fixed
+- Auth failure in /editorial now redirected to /users/sign_in
 
 
 ## [v1.3.0] - 2016-09-07
--- a/app/controllers/admin/application_controller.rb
+++ b/app/controllers/admin/application_controller.rb
@@ -7,6 +7,7 @@
 module Admin
   class ApplicationController < Administrate::ApplicationController
     include ::IncompleteTfaSetup
+    include ::AccessDeniedConcern
 
     before_action ->() { authorize! :manage, :all }
     before_action :complete_two_factor_setup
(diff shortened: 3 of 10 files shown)


app/controllers/editorial/editorial_controller.rb
```diff
@@ -9,6 +9,13 @@ class EditorialController < ::ApplicationController
 
     layout 'editorial'
 
+    rescue_from CanCan::AccessDenied do |exception|
```
tessereth: You only want to redirect to sign in if the user isn't signed in already

freshtonic: Looks like we'll need to fix this in two other places (application controller and admin controller) too.

I'm also getting concerned about the lack of DRYness in our controllers. We have 3 controller hierarchies :/

tessereth:

freshtonic:
@tessereth updated changelog

app/controllers/editorial/editorial_controller.rb
```diff
@@ -9,6 +9,8 @@ class EditorialController < ::ApplicationController
 
     layout 'editorial'
 
+    include AccessDeniedConcern
```
freshtonic: Yep. I think it's leftover from refactoring. I'll remove it.

app/views/errors/unauthorized.html.haml
```diff
@@ -0,0 +1,8 @@
+= content_for :title do
+  You are not authorised to access this page (401)
+
+%article.content-main.callout--calendar-event
+  = render 'application/alerts'
```
tessereth: Does this use the application layout? Because if so, that layout already renders the alerts.

tessereth:
You added an empty `public/401.html` file. Is that required?

tessereth:
This does not appear to be modifying the admin application_controller. Should it be?

freshtonic:
@tessereth I've considered admin out of scope for this.

tessereth:
LGTM

app/views/errors/unauthorized.html.haml
```diff
@@ -0,0 +1,7 @@
+= content_for :title do
+  You are not authorised to access this page (401)
```
bjgordon: authori**z**ed or authori**s**ed? With this change we now have both. Since we're .au we should consistently use **s**, so I suggest we change to that. (in a new card if you dont want to do it here)

bjgordon:
LGTM

PR #479, opened 2016-09-05
SITES-680 Move GTM to body

bjgordon:
GA is supposed to be put in `<head>`, but GTM is supposed to go in `<body>`. I think I originally did this wrong as during implementation of GTM I was experimenting with both GA and GTM trying to get turbolinks working. The GTM `<noscript>` tag is probably ignored when in `<head>`, so it's worth fixing.


maxious:
LGTM

tessereth:
LGTM

PR #478, opened 2016-09-05
[SITES-567] Implement content links using liquid templates

tessereth:
Entering the content body
```
{% assign careers = nodes['a183986f-5191-46d4-8cc8-d8bf30ab055b'] %}
{% assign people = nodes['eca9b7c8-6774-4ab3-954d-be3da0c654ed'] %}

## The following pages are in this section

### {{ careers.link_to }}

{{ careers.short_summary }}

### {{ people.link_to }}

{{ people.short_summary }}
```
Results in a published page of
image

Entering invalid content results in an error being displayed on the submission page and the error being highlighted in red.
```
Link 1: {{ nodes['bad'].link_to }}  
Link 2: {{ nodes['bad2'].link_to }}
```
image

Note: As it currently stands, getting the UUID of a node is rather a pain. We should probably have a widget on the submission form that lets you search by name/path. But this PR is big enough as it is.

PR #481, opened 2016-09-05
News page update [GOVAU-422]

simonschwartz:
This PR updates the layout of the news article page to match the agency teams mvp design. It also changes the 'Published by' link to link to the department/minister news page rather than the department/minister landing page.

This PR also includes styling for related articles block and contact card. 

Related articles aren't intended to be functional in the platform quite yet. But the plan is to see what is possible in the future.

Authoring the contact card information isn't intended to be supported in the platform yet as the work that agency is doing on 'People and structure' will probably affect how this information should be handled in the platform.

news-article


ltankey:
LGTM

app/views/templates/news_article.html.haml
```diff
@@ -3,26 +3,30 @@
...
+    %span
+      Published by
+    %a= link_to(node.section.name, section_news_articles_path(section: @section.home_node.slug))
 
 = render partial: 'templates/content_body', object: node.content_body
 
 =@hierarchy
+
+-#%div.list-horizontal
+  = render partial: 'templates/related_sections', locals: {node: node}
```
tessereth: I remember needing the `list-horizontal` class to get the tags to render correctly. Plus, this relegates the related sections to the very end of the document when I would have thought it should be near the "published by" entry. Has this been tested and a decision made to put them here?

tessereth:
Just a note, this will have merge conflicts with #478

app/views/templates/news_article.html.haml
simonschwartz: Tags are out for MVP of this page

Having related articles at the end of the document makes sense to me as it seems that users would naturally be looking for related articles after reading the content. In usability testing sessions related articles were placed in this position and tested well on desktop and mobile. 

I would say for MVP this is a fine approach but its an interesting question. I'll let Sam know and we might even test this idea next week in the prototype.

Any opinions @klepas

ltankey: How are related news articles defined?

maadonna: They haven't been defined...

I'm not sure why the 'tags'  (which were topics) were dropped. It is important for people to be able to navigate from this news story to the topics that it relates to. I'll chase it up.

ltankey: I thought they were still in for the last prototype - and they're still in prod at the moment. It'd be great to get clarity on this, thanks @maadonna.

tessereth: I agree with you re related news articles but the `related_sections` template here is not rendering related news articles. It's rendering the additional sections (departments, ministers etc) that contributed to the news article.
image
I'm wondering where this should go on the page. Note: the keyword 'Tags' here was made up by me but really it's just the list of related sections.

maadonna: Yeah, this always should have been 'related sections that are topics, not that are ministers or agencies' (as the purpose was to help people navigate to the relevant topics. But I don't know that was ever discussed/specified properly even on the original implementation.

simonschwartz: sorry @maadonna dropped was the wrong word. Left out until the Agency team has a clearer idea of how/where to represent them is what I should have said

simonschwartz: @ltankey We aren't sure how to define related articles yet. We are leaving out related news articles from the platform for now because there needs to be a conversation with the platform people about what is possible for an MVP. One thing that came up in usability testing was that if the related articles weren't meaningfully related then it would be super annoying. Could just be manually entered in the article editor, could be rails magic etc

michaelhugill:
Hi, just letting you know we should be able to bring you some clarity today. Will get back to you as soon as I can.

michaelhugill:
Hi again, @samcourt and I will look at this tomorrow and make updates to the JIRA ticket and share here.

tessereth:

app/decorators/news_article_decorator.rb
```diff
@@ -6,4 +6,16 @@ def related_sections
       related_section == object.section
     end
   end
+
+  def publishers
```
ltankey: This method creates confusion with `NewsArticle.publisher` - is this field no longer required if we instead use the distribution list?

app/views/templates/_related_news_articles.html.haml
```diff
@@ -0,0 +1,12 @@
...
+  %ul
+    %li
+      %a{:href => "#"}
+        A related article here
+    %li
+      %a{:href => "#"}
+        A related article here
+    %li
+      %a{:href => "#"}
+        A related article here
```
ltankey: We can probably make this real content, if we base this on the last 5 news articles in chronological order from the associated topics.

We have already implemented this kind of query for news filtering - it might be possible to abstract the code that was written there to take advantage of here.

CHANGELOG.md
```diff
@@ -10,6 +10,7 @@ This file is influenced by http://keepachangelog.com/.
 ### Changed
 - Incorrect attempts at verifying identity locks the user out
 - Change git merge strategy for CHANGELOG.md to reduce conflicts
+- Update News article page template
```
ltankey: It might be worth including what was changed - i.e. "including publishers and related news articles and topics"

simonschwartz:
Based on discussion between @maadonna @samcourt and @AlzbetaT I have updated the news article page. 

These changes include: 
- remove 'Back to news' button
- make department link go to department landing page
- allow multiple publishers to be displayed [SITES-706]
- include 'Related topics:' in the article information section
- make 'Related to:' links only include topics [SITES-705]
- ensure the news article snippet displays the article metadata correctly

screenshot 4

news-story-page

tessereth:
Changelog still needs to be updated to move the change into the unreleased section, otherwise LGTM

app/views/layouts/news_article.html.haml
```diff
@@ -1,20 +1,17 @@
...
-  = content_for :after_header do
-    %div{class: "hero-sml #{hero_class(@section)}"}
-      .wrapper
-        %p
-          = @section.name
-
   = content_for :toolbar_actions do
     = render partial: 'application/page_actions', locals: {section: @section, node: @node}
 
   = content_for :meta_description do
```
ltankey: This line looks redundant.

ltankey:
LGTM

PR #482, opened 2016-09-06
SITES-630 Timeout sessions after 90 minutes idle

maxious:
Use built-in devise functionality for this.


ltankey:
LGTM

tessereth:
LGTM though if you can figure out a way to test it, that would be great.

maxious:
Hmm devise has https://github.com/plataformatec/devise/blob/master/test/models/timeoutable_test.rb

tessereth:
Fair call. Makes sense.

maxious:
rebased

PR #483, opened 2016-09-06
WIP SITES-685 Performance on news#index

bjgordon:
Still need to improve behaviour of `related_sections`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -28,6 +28,7 @@ This file is influenced by http://keepachangelog.com/.
 - GTM tag is now in <body>
 - email address is now validated on feedback form
 - removed prompt to sign up when unauthenticated
+- Performance of /news
 
 ## [v1.2.1] - 2016-08-31
 ### Added
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ This file is influenced by http://keepachangelog.com/.
 ### Changed
 
 ### Fixed
+- Performance of /news
 
 
 ## [v1.3.0] - 2016-09-07
@@ -28,7 +29,6 @@ This file is influenced by http://keepachangelog.com/.
 - GTM tag is now in <body>
 - email address is now validated on feedback form
 - removed prompt to sign up when unauthenticated
-- Performance of /news
 
 ## [v1.2.1] - 2016-08-31
 ### Added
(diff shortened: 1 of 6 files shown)


bjgordon:
The page time is improved but still not that great - locally down from around 4.5s to 2.5s using a prod db dump.
In `/newrelic` this change has reduced the number of database queries of `/news` request from 457 to 336. The remainder are little queries for each related_sections partial, but I havent been able to reduce that. Open to ideas.

freshtonic:
@bjgordon this is using production Rails config locally right?

bjgordon:
@freshtonic actually no. I was just interested in relativele performance change so didnt think itd matter... ?

freshtonic:
@bjgordon benchmarking in development mode is not going to tell you anything useful IMHO.  The overheads are quite large & some internal caching is disabled.

freshtonic:
@bjgordon basically, I wouldn't count on dev mode to have the same relative performance change characteristics as prod mode.

bjgordon:
@freshtonic Ok I'll test again in prod mode.

bjgordon:
In production mode before the change it takes around 1.5s with 180 db hits, after the change around 1s with 80 db hits.

tessereth:
Should this same change be applied to the `/:section/news` pages as well?

freshtonic:
Is something hitting that `news_articles` helper in a loop to load the data?  80 database hits is way too many. "around 5 to 7" database hits to build a page is a reasonable number. That's normally enough to load the user & roles and a the resources to load the page.

bjgordon:
@freshtonic I think the issue is in `_related_sections.html.haml`, each news_article's related_sections and home_node is two separate queries.

freshtonic:
@bjgordon so the solution to this is to front load everything we need and not invoke helpers to load additional data. Judicious use of `:includes` is key.

tessereth:

bjgordon:
Have tried moving related_sections to the model, but can't `.includes` it as it's instance dependent. Also tried getting rid of the use of the decorator, but that increased the db hits as filtering is now done in db rather than in memory, and didn't help with preloading, so should probably revert that. Handing this over to @freshtonic to take a look at.

freshtonic:
Replaced by https://github.com/AusDTO/gov-au-beta/pull/491

PR #490, opened 2016-09-08
[SITES-366 SITES-665] Feature/news filter by minister

ltankey:
image


app/controllers/news_controller.rb
```diff
@@ -1,22 +1,39 @@
...
+        @section = SectionHome.find_by!(slug: params[:section]).section
+      rescue ActiveRecord::RecordNotFound => e
+        e.instance_variable_set(:@_rollbar_do_not_report, true)
+        raise
+      end
+
       set_menu_nodes
-      @articles = news_articles section: @section
+      @filters = [@section]
+      @articles = news_articles#(@section)
```
tessereth: commented out code?

app/controllers/news_controller.rb
```diff
@@ -1,22 +1,32 @@
...
+      @articles = news_articles
     else
       @articles = news_articles
+
+      if @filters.present?
+        @articles = @articles
+                        .includes(:sections)
+                        .references(:sections)
+                        .where('sections.id IN (?)', @filters.collect(&:id))
+      end
```
freshtonic: Preloading all the associations needed for the view like in the #491 is the key.

```ruby

File: app/controllers/news_controller.rb [hunk 3]
```diff
@@ -37,13 +47,33 @@ def show
...
+  # with section ids
+  def set_filters
+    @filters = []
+
+    [:ministers, :departments].each do |filter_type|
+      if params[filter_type].present?
+        id_list = params[filter_type].map(&:to_i)
+
+        id_list.reject! do |item|
+          !Section.exists?(item)
```
freshtonic: This will incur a DB round trip every run through the loop. Instead, preload all the section IDs into an array and check that.

app/helpers/news_helper.rb
```diff
@@ -1,17 +1,21 @@
 module NewsHelper
-  def news_article_class
+  def news_articles#(section: nil)
```
freshtonic: This will conflict with the changes I have made in the other PR. Just a heads up :)

app/controllers/news_controller.rb
ltankey: Good catch!

app/controllers/news_controller.rb
ltankey: Would it suffice to rewrite the declaration of line 23 with

`@articles = news_articles.`

This would preload the associations via `news_helper.rb`.

ltankey: As `@articles` uses the `news_articles` on line 20, does this suffice? The subsequent calls merely extends the query.

app/controllers/news_controller.rb
```diff
@@ -36,13 +45,46 @@ def show
...
+        raise
+      end
+    end
+  end
+
+
+  def filtered_articles
+    if @filters.present?
+      preload_news_articles
+        .includes(:sections)
```
freshtonic: Pretty sure you don't need this line - it's already included. You can keep the `references` and extra `where` clause though.

freshtonic:
Just one comment to address, otherwise LGTM

app/controllers/news_controller.rb
ltankey: If includes is removed, an error is returned for the tests:

`PG::UndefinedTable: ERROR:  missing FROM-clause entry for table "sections"`

app/controllers/news_controller.rb
```diff
@@ -36,13 +40,46 @@ def show
 
 
   private
+  # An array of sections the user is filtering by, loaded by params
+  # with section ids
+  def set_filters
+    @filters = []
+    section_ids = Section.all.collect(&:id)
```
tessereth: I think `Section.all.collect(&:Id)` will load all sections from the database. You could probably do `Section.pluck(:id)` instead to run `SELECT id FROM sections`.

app/helpers/news_helper.rb
```diff
@@ -16,11 +7,22 @@ def preload_news_articles(query_root = NewsArticle)
...
       .published
       .by_release_date
       .by_published_at
       .by_name
   end
+
+
+  # List of Ministers for filtering
+  def ministers_list
+    Minister.order(:name).decorate.sort_by! { |minister| minister.sort_order }
```
tessereth: You might want to preload the section home so you can get the name?

app/views/news/_filter_section.html.haml
```diff
@@ -0,0 +1,11 @@
+%details{open: nil, data: {label: "#{section_type}"}, aria: {expanded: true}}
+  %summary= section_title
+  .accordion-panel{style: "overflow: scroll; max-height: 350px;"}
+    %fieldset
+      - sections.each do |s|
+        - if @filters.present?
+          - checked = @filters.include?(s)
+
+        = check_box_tag "#{section_type}[]", s.id, checked, id: "#{section_type}_#{s.id}"
```
tessereth: commented out code?

app/controllers/news_controller.rb
freshtonic: Then the `preload_news_articles` method should be fixed so that it can be referenced.

spec/controllers/news_controller_spec.rb
```diff
@@ -67,4 +67,82 @@
...
+    let!(:department_b) { Fabricate(:department) }
+    let!(:minister_c) { Fabricate(:minister) }
+    let!(:minister_d) { Fabricate(:minister) }
+
+    let!(:a_home_node) { Fabricate(:section_home, section: department_a) }
+    let!(:b_home_node) { Fabricate(:section_home, section: department_b) }
+    let!(:c_home_node) { Fabricate(:section_home, section: minister_c) }
+    let!(:d_home_node) { Fabricate(:section_home, section: minister_d) }
+
+    let!(:article_a) { Fabricate(:news_article, state: 'published', section: department_a, sections: [department_a]) }
```
tessereth: Did you want to update the fabricator to automatically add the section to the sections array?

tessereth:
I left a couple of comments but LGTM

spec/controllers/news_controller_spec.rb
ltankey: Given the tears shed in working this out, I think it's my duty to prevent it from happening to anyone else ever again...

app/helpers/news_helper.rb
ltankey: This is a good point, however we should bundle it into a performance tech-debt card.

PR #491, opened 2016-09-08
SITES-685 improve news page performance

freshtonic:


CHANGELOG.md
```diff
@@ -10,6 +10,7 @@ This file is influenced by http://keepachangelog.com/.
 ### Fixed
 - Pluralise 'ministers' only where appropriate
 - Fixed HSTS settings so we don't force every gov.au subdomain to run over https
+- Perforance of /news page
```
ltankey: Typo: just missing an `m` 😃

freshtonic: LOL

bjgordon:
Nice, LGTM

app/helpers/news_helper.rb
```diff
@@ -8,10 +8,18 @@ def news_article_class
...
-    else
-      NewsArticle.published.by_release_date.by_published_at
-    end
+    query_root = if section.present?
+                   section.news_articles
+                 else
+                   NewsArticle
+                 end
+    query_root
+      .preload([
```
coderdan: I hadn't actually seen preload. Is it new in Rails 5?

app/helpers/news_helper.rb
```diff
@@ -8,10 +8,18 @@ def news_article_class
   end
 
   def news_articles(section: nil)
-    if section.present?
-      section.news_articles.published.by_release_date.by_published_at
-    else
-      NewsArticle.published.by_release_date.by_published_at
-    end
+    query_root = if section.present?
```
coderdan: Could you make query_root a method to keep this method a bit simpler?

app/helpers/news_helper.rb
freshtonic: I only just came across it. AFAIK it does exactly the same as `includes` but I think the name is more intention-revealing. From the docs:

app/helpers/news_helper.rb
freshtonic: Done. See latest commit.

ltankey:
LGTM

app/helpers/news_helper.rb
```diff
@@ -7,11 +7,20 @@ def news_article_class
...
+  def preload_news_articles(query_root = NewsArticle)
+    query_root
+      .preload([
+        {:section  => {:home_node => :parent}},
+        {:sections => {:home_node => :parent}}
+      ])
+      .published
+      .by_release_date
+      .by_published_at
+      .by_name
```
ltankey: is by_name necessary? Published_at is a timestamp - is this just for edge cases?

app/controllers/news_controller.rb
```diff
@@ -13,9 +11,9 @@ def index
     if params[:section].present?
       set_section
       set_menu_nodes
-      @articles = news_articles section: @section
+      @articles = preload_news_articles(@section.news_articles)
```
ltankey: Do we need to pass `@section`? If the helper is included, it would have access to it by default.

ltankey:
minor comments aside, LGTM

app/helpers/news_helper.rb
freshtonic: There was an existing spec that tests for order by dates then name. The previous implementation did not sort by name and it was accidentally passing. So I made it sort by name.

PR #494, opened 2016-09-11
feature: Basic image upload to S3

maxious:
Addresses SITES-678

This implementation uploads files to S3 with the SHA1 in the filename suffixed with a "style" - original, medium and thumbnail. That is, thumbnails and downscaled images are automatically generated during the file upload. However, width/height are not stored in the database.

The Asset.asset_file object has a .url function. For testing this is generating raw S3 URLs but can be customised as "s3_host_alias"

An asset can be uploaded through editorial. There is no metadata entered manually and the URL is displayed in a flash message after upload.


config/routes.rb
```diff
@@ -34,6 +34,7 @@
   resource :feedback, controller: 'feedback'
 
   namespace :editorial do
+    resources :assets, only: [:index, :new, :create]
```
ltankey: There's no corresponding action for `:index` in the controller.

ltankey:
We should probably update the changelog also.

config/routes.rb
maxious: I just need the template to render. is it good practice to make a blank placeholder?

ltankey: I thought so, but as it turns out convention says if the controller exists that's all that's required.

ltankey:
LGTM

app/forms/asset_form.rb
```diff
@@ -0,0 +1,5 @@
+class AssetForm < Reform::Form
+  include Reform::Form::ActiveRecord
```
tessereth: What's the reason for this?

app/controllers/editorial/assets_controller.rb
```diff
@@ -0,0 +1,32 @@
...
+    def new
+      @form = AssetForm.new(Asset.new)
+    end
+
+    def create
+      @form = AssetForm.new(Asset.new)
+      if @form.validate(asset_params)
+        @form.sync
+        asset = @form.model
+        asset.uploader_id = current_user.id
```
tessereth: I think it might be more idiomatic to say `asset.uploader = current_user`

app/views/editorial/assets/index.html.haml
```diff
@@ -0,0 +1,8 @@
+= content_for :title do
+  = "Assets"
+
+- breadcrumb :new_asset
```
tessereth: This looks like the wrong breadcrumb

config/breadcrumbs.rb
```diff
@@ -99,6 +99,11 @@
   parent :editorial_root
 end
 
+crumb :new_asset do
+  link 'Create asset', new_editorial_asset_path
+  parent :editorial_root
```
tessereth: It may or may not make more sense to use the asset index page as the parent for the new asset page.

db/schema.rb
```diff
@@ -10,11 +10,23 @@
...
+    t.integer  "uploader_id"
+    t.string   "asset_file_file_name"
+    t.string   "asset_file_content_type"
+    t.integer  "asset_file_file_size"
+    t.datetime "asset_file_updated_at"
+    t.string   "asset_file_meta"
+    t.string   "asset_file_fingerprint"
+    t.datetime "created_at"
+    t.datetime "updated_at"
+  end
```
tessereth: We probably need a FK constraint on the uploader. And possibly a unique constraint on the file name if we're going to use that as our identifier.

tessereth:
There are no controller tests for this PR

db/schema.rb
ltankey: Although we are soft-deleting, we probably don't want assets deleted if an associated user-account is, particularly if they're shared.

tessereth: The same applies to submissions etc. That's exactly why we're soft deleting users. So we should be okay having the foreign key constraint.

tessereth:
Other than the breadcrumb for the asset index page, this LGTM.

PR #496, opened 2016-09-13
stop window from jumping down upon checkbox selection in overflo…

ltankey:
This bug affects checkboxes in accordions with overflow.

The impact would be that for checkboxes far down in the accordion, a selection would make the screen jump down to where the checkbox would exist without the overflow on that area.

This is technically a UIKIT bug, but we should be pragmatic and merge this in advance of an update from them.


maxious:
LGTM

bjgordon:
LGTM

PR #497, opened 2016-09-13
SITES-701 improve HTTP caching

freshtonic:


app/controllers/editorial/news_controller.rb
```diff
@@ -6,7 +6,7 @@ class NewsController < EditorialController
 
     def index
       # TODO: return news editorial index
-      bustable_fresh_when(@sections)
+      with_caching(@sections)
```
maxious: Does this change the etag when news articles are added?

freshtonic: No, but from reading the code it does't render any news articles on the editorial news index so it shouldn't matter.

freshtonic: That particular invocation will only change the etag based on `@section.updated_at`.

maxious:
LGTM + this seems to be working in a way that makes sense now

ltankey:
LGTM

PR #499, opened 2016-09-13
SITES-611 Add beta invite system

bjgordon:
Adds a toggle to require all visitors to have an invite. Sorry this PR ended up a bit large. 

When toggled on, all unknown visitors get redirected to
image

Logged in users can still view the site as normal, i.e. editors and admins dont need an invite.

Any logged in user can send an invite:
image

They get an email:
image

And when they accept it:
image


.env.sample
```diff
@@ -21,4 +21,7 @@ PORT=3000
 # Use the following to override if necessary.
 # FORCE_2FA=true
 # DISABLE_2FA=true
-# TFA_SECRET_KEY
```
tessereth: We shouldn't remove this line

bjgordon: No we shouldn't, good pick up

app/dashboards/invite_dashboard.rb
```diff
@@ -0,0 +1,67 @@
...
+
+  # FORM_ATTRIBUTES
+  # an array of attributes that will be displayed
+  # on the model's form (`new` and `edit`) pages.
+  FORM_ATTRIBUTES = [
+    :code,
+    :accepted_token,
+    :accepted_at,
+    :email,
+  ].freeze
```
tessereth: I'm not sure what the value of editing this model is

app/forms/invite_form.rb
```diff
@@ -0,0 +1,6 @@
+class InviteForm < Reform::Form
+  property :email
+  validates :email, :presence => true
+  validates :email, :email => true
+  validates_uniqueness_of :email
```
tessereth: I can see a use case where someone needs more than one invite. Maybe leave this out for now?

app/services/invite_creator.rb
```diff
@@ -0,0 +1,25 @@
...
+      Notifier.beta_invite(invite).deliver_later
+    end
+  end
+
+  private
+  def generate_code
+    candidate = SecureRandom.hex(16)
+    if Invite.exists?(code: candidate)
+      # Collision? Try again.
+      generate_code
```
tessereth: Best to have only a finite number of retries. No one likes a stack overflow.

app/views/invites/new.html.haml
```diff
@@ -0,0 +1,13 @@
...
+
+- breadcrumb :invite
+
+%h1
+  Send an invite
+
+%section#content.content-main.form_wrapper{tabindex: -1}
+  = simple_form_for(@form) do |f|
+    = f.input :email, as: :email
+    = f.button :button, 'Send invite'
```
tessereth: The `%h1` should be inside the `#content` block

config/breadcrumbs.rb
```diff
@@ -28,6 +28,10 @@
   parent :public_node, Node.root_node
 end
 
+crumb :invite do |name|
+  parent :public_node, Node.root_node
```
tessereth: Should also have a link entry. Something like `link 'Invite', invite_path`.

app/services/invite_creator.rb
freshtonic: 16^16 = 18,446,744,073,709,551,616 combinations. If it recurses even once in a thousand years it would be an epic fluke 😝

db/migrate/20160822111550_create_invites.rb
```diff
@@ -0,0 +1,13 @@
...
+  def change
+    create_table :invites do |t|
+      t.string :code
+      t.string :accepted_token
+      t.timestamp :accepted_at
+      t.string :email
+      t.timestamps
+    end
+    add_index :invites, :code
+    add_index :invites, :accepted_token
```
tessereth: Do these need to be unique?

app/controllers/invites_controller.rb
```diff
@@ -0,0 +1,64 @@
...
+
+      flash[:alert] = "Invite has been sent to #{@invite.email}"
+      redirect_to new_invite_url
+    else
+      render :new
+    end
+  end
+
+  def show
+    @invite = Invite.find_by!(:code => params[:id])
```
ltankey: Perhaps we could standardise, as code field is confusing given usage of `token` elsewhere.

app/controllers/invites_controller.rb
```diff
@@ -0,0 +1,64 @@
+class InvitesController < ApplicationController
+  skip_before_action :authorize_invite_access
+
+  before_action :check_invite_required, only: [:required, :show]
+
+  load_and_authorize_resource :only => [:new, :create]
+
+  layout :resolve_layout
+  def required
```
ltankey: This isn't necessary - Rails by convention will route directly to a named view.

app/forms/invite_form.rb
ltankey: I think these three validation lines can be condensed to one:

`validates :email, :uniqueness => true, :presence => true, :email => true`

app/models/ability.rb
```diff
@@ -24,6 +24,8 @@ def initialize(user)
       # TODO: Should all logged in users be able to view editorial pages?
       can :view, :editorial_page
 
+      can :create, Invite
```
ltankey: Should all existing users be able to invite others? Or should this be an admin function.

db/migrate/20160822111550_create_invites.rb
bjgordon: I've added a new migration to make them unique

app/models/ability.rb
bjgordon: I've changed so it's admin only

app/dashboards/invite_dashboard.rb
bjgordon: Changed to read only

app/services/invite_creator.rb
bjgordon: Even worse it's 256^16. It seems like a very low risk...

app/controllers/invites_controller.rb
bjgordon: The  invite `code` is emailed to the user, then  `accepted_token` is generated and put in a cookie when they accept. Open to suggestions on other names.

PR #501, opened 2016-09-14
Make curl return non-zero exit code on a 404

bjgordon:


circle.yml
```diff
@@ -34,7 +34,7 @@ deployment:
       - cf auth $CF_USER_PROD $CF_PASSWORD_PROD
       - cf target -o $CF_ORG_PROD
       - cf target -s $CF_SPACE_PROD
-      - curl https://api.rollbar.com/api/1/deploy/ -F access_token=$ROLLBAR_PROD_ACCESS_TOKEN
+      - curl -f https://api.rollbar.com/api/1/deploy/ -F access_token=$ROLLBAR_PROD_ACCESS_TOKEN
```
tessereth: If rollbar is unavailable, we probably still want to deploy

bjgordon: Ok I've removed

tessereth:
LGTM

PR #503, opened 2016-09-16
SITES-673 Fix database and AWS creds in db.rake

bjgordon:
WIP as waiting on S3 creds to test...


bjgordon:
Have tested this with my own S3 creds. rake db:s3:upload worked!

PR #505, opened 2016-09-19
SITES-681 Image asset listing and liquid image tag

maxious:


config/initializers/kaminari_config.rb
```diff
@@ -0,0 +1,10 @@
+Kaminari.configure do |config|
+  # config.default_per_page = 25
```
ltankey: Do we need any of these default configuration items?

app/controllers/editorial/assets_controller.rb
```diff
@@ -18,8 +18,8 @@ def create
         asset = @form.model
         asset.uploader = current_user
         if asset.save
-          flash[:notice] = "Asset created, available at https:#{asset.asset_file.url}"
-          redirect_to editorial_root_path and return
+          flash[:notice] = "Asset successfully created"
+          redirect_to editorial_assets_path and return
```
ltankey: The tests probably need to be updated to reflect this.

config/initializers/kaminari_config.rb
maxious: Nope. You can generate the config file later but I thought better to have something you can grep for.

app/controllers/editorial/assets_controller.rb
maxious: Done.

PR #516, opened 2016-09-29
feat: remove categories and news links from homepage

ltankey:
image

### DIFF
--- a/app/views/templates/root_node.html.haml
+++ b/app/views/templates/root_node.html.haml
@@ -2,18 +2,18 @@
 
 %article.content-listing#content{tabindex: -1}
 
-  %section.home-category-list
-    %h2
-      Information and services
-
-      %ul.list-vertical--thirds.real-list
-        - @categories.each do |category|
-          %li
-            %article
-              %h3
-                = link_to(category.name, category_path(category.slug))
-              %p
-                = category.short_summary
+  -#%section.home-category-list
+  -#  %h2
+  -#    Information and services
+  -#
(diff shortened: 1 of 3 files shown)


app/views/templates/root_node.html.haml
```diff
@@ -2,18 +2,18 @@
...
-
-      %ul.list-vertical--thirds.real-list
-        - @categories.each do |category|
-          %li
-            %article
-              %h3
-                = link_to(category.name, category_path(category.slug))
-              %p
-                = category.short_summary
+  -#%section.home-category-list
```
coderdan: Is it better to use HAML comments? i.e.

    /%section.home...

ltankey: I think the main difference is that by using `/%section` syntax is that the template block will render out as a HTML comment, i.e.:

```
        <!-- %section.home-category-list -->
        <!-- %h2 -->
        <!-- Information and services -->
        <!-- %ul.list-vertical--thirds.real-list -->
        <!-- - @categories.each do |category| -->
        <!-- %li -->
        <!-- %article -->
        <!-- %h3 -->
        <!-- = link_to(category.name, category_path(category.slug)) -->
        <!-- %p -->
        <!-- = category.short_summary -->
```

whereas the syntax included in this PR will not include the commented blocks at all.

Check out the source of https://pr-516-gov-au-beta.apps.staging.digital.gov.au/

coderdan: OK cool. That makes sense.

PR #519, opened 2016-10-01
ref: change various parts of the homepage copy

ltankey:
Card 29.

image

### DIFF
--- a/app/views/templates/root_node.html.haml
+++ b/app/views/templates/root_node.html.haml
@@ -84,7 +84,7 @@
     %h2.home-heading.small Local government
     %p
       The 6 states and the Northern Territory have established 1 further
-      level of Government. Local governments (also known as city, shire, town or
+      level of government. Local governments (also known as city, shire, town or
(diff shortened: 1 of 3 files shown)


app/views/templates/root_node.html.haml
```diff
@@ -16,77 +16,75 @@
...
+            %p A secure myGov account lets you link a range of Australian Government services with one username and password, all in one place.
     .state-government
-      %h3.home-heading.small State and territory governments
+      %h2.home-heading.small State and territory governments
       %ul.list-vertical
         %li.nsw
           %span.image
           %article
             %h3.state-title New South Wales
-            %a{href: 'https://www.nsw.gov.au/'} www.nsw.gov.au
```
coderdan: I admit that www looks old and clunky but the URL is technically www.nsw.gov.au if someone doesn't follow the link and type the URL into the browser we should make sure that they all redirect.

coderdan: And actually on further inspection this looks to be the case.

PR #522, opened 2016-10-05
Fix/ui tweeks

TrebBrennan:
Basic updates to the responsiveness and layout of the homepage

screencapture-gov-au-beta-apps-staging-digital-gov-au-1475632491986


app/views/layouts/root.html.haml
```diff
@@ -2,21 +2,21 @@
   = yield
 
   = content_for :header do
-    %section.prerelease-govau--header.home
+    %section
+      //.prerelease-govau--header.home
```
ltankey: If we use the syntax `-#` instead of `//` the HAML won't be rendered as HTML comments, i.e.: 
`<!-- /.prerelease-govau--header.home -->`

PR #526, opened 2016-10-05
Fix/browser-support-edits

TrebBrennan:
Minor improvements for layout on older browsers.


TrebBrennan:
Thanks @coderdan, this is just WIP at the moment. I needed the QA-Fire URL to test in Browserstack so this PR is purely just to get something I get look at until I solve this issues :(

PR #539, opened 2016-10-07
ref: simplify feedback form

ltankey:
image


app/forms/feedback_form.rb
```diff
@@ -1,10 +1,10 @@
 class FeedbackForm < Reform::Form
   property :url
-  property :email
   property :comment
-  property :role
-  property :organisation
-
-  validates :email, :comment, :role, :organisation, presence: true
-  validates :email, :email => true
+  # property :email
```
coderdan: Should we just remove these?

PR #542, opened 2016-10-07
Ensure GA tracking is called asynchronously

coderdan:
This approach guarantees that google analytics calls don't get made before everything has loaded. It also allows the analytics source code to be placed in the asset pipeline.

### DIFF
--- a/app/views/layouts/application.html.haml
+++ b/app/views/layouts/application.html.haml
@@ -12,7 +12,7 @@
 <html lang="en-AU" class="no-js"><!--<![endif]-->
 
 = render 'application/head',  title: (content_for?(:title) ? "#{yield(:title)} | GOV.AU" : 'GOV.AU' )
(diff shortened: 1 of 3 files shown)


app/views/layouts/application.html.haml
```diff
@@ -12,7 +12,7 @@
 <html lang="en-AU" class="no-js"><!--<![endif]-->
 
 = render 'application/head',  title: (content_for?(:title) ? "#{yield(:title)} | GOV.AU" : 'GOV.AU' )
-%body
+%body{data: { 'ga-id' => ENV['GA_TRACKING_ID'] }}
```
ltankey: Do we need to care about what happens if no tracking id is provided?

coderdan: It will just be null and the JS code should handle it

PR #546, opened 2016-10-10
Fix for sticky footer

TrebBrennan:
Basic flex-box addition for `role=main` because UI-Kit uses just `main`

### DIFF
--- a/app/assets/stylesheets/_content_block.scss
+++ b/app/assets/stylesheets/_content_block.scss
@@ -3,3 +3,8 @@
   border-style: solid;
   border-width: 1px;
 }
+
+//Fix for sticky footer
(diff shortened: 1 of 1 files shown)


app/assets/stylesheets/_content_block.scss
```diff
@@ -2,4 +2,9 @@
...
   border-style: solid;
   border-width: 1px;
-}
\ No newline at end of file
+}
+
+//Fix for sticky footer
+div[role='main'] {
+  flex: 1 100%;
+}
```
ltankey: We should probably look at shifting this to a better home when we refactor, but as this is commented it should make it obvious where an appropriate home is.

PR #550, opened 2016-10-11
Fixed favicon

coderdan:


coderdan:
I'm treating it as a hotfix. We can backmerge into develop (ala a gitflow hotfix) @ltankey

ltankey:
All good I approved it

PR #568, opened 2016-11-03
use kramdown renderer instead of redcarpet

ltankey:
This PR uses Kramdown instead of RedCarpet as the markdown renderer.

This gives authors the ability to use definition lists.

Due to the way that Kramdown compiles its documents, table-of-contents are refactored out as a separate explicit template specification, and are instead compiled as part of the content itself. An inline specification is used to inject the TOC at the head of the document's contents.

Note: various content-tests have been refactored to suit the kramdown markdown syntax for headings, which requires that a heading is preceded by a blank line.


app/services/markdown_renderer.rb
```diff
@@ -1,25 +1,40 @@
...
   def render(renderer, markdown_content)
     unless markdown_content.nil?
-      raw = renderer.render(markdown_content)
+
+      # Prepend the TOC markdown so that kramdown expands
+      # it at the top of the document. We should always inject this, as the
+      # @toc_depth indicates the range  of the TOC which defaults to rendering
+      # nothing.
+      # http://kramdown.gettalong.org/converter/html.html#toc
+      toc_header = "* TOC\n{:toc}\n\n"
```
coderdan: A small thing but maybe worth setting this as a constant?

app/views/templates/_toc.html.haml
```diff
@@ -1,8 +1,11 @@
+-# DEPRECATED - TOC is generated as part of content
```
coderdan: Can we remove entirely now?

config/initializers/kramdown.rb
```diff
@@ -0,0 +1,89 @@
+module Kramdown
```
coderdan, line 1: Is this all boilerplate or so I need to review?

app/services/markdown_renderer.rb
ltankey: Good point!

coderdan (commented):
Is it worth adding/updating specs for the node decorator and the markdown renderer?

app/views/templates/_toc.html.haml
ltankey: Done - will be in our commit history if we ever want something similar to this.

config/initializers/kramdown.rb
ltankey: The `nav_wrapped` method is the only thing that isn't boilerplate.

ltankey:
Fair question - the specs have been updated to work with these changes, but the output hasn't changed.

I could add some additional tests for the `node_decorator` however there is no specs for the markdown renderer service, which might be worth adding.

coderdan:
Yeah up to you but I think the specs for the markdown renderer would be a good improvement.

PR #570, opened 2017-11-28
Cloud migration

m-richo:
getting the site ready to deploy to new Cloud Foundry environments.
- new build packs require newer ruby (2.3.1 -> 2.3.5)
- refresh gem deps in `Gemfile.lock` (particularly rubyzip which has a known vulnerability)
- update CF deployment to use new zero-down-time plugin
- Update Prometheus gem as there were errors (perhaps wasn't necessary in hindsight)
- correct prometheus config
- comment out 1 failing test (could do with a rails developer to review)


circle.yml
```diff
@@ -15,26 +16,28 @@ test:
...
-      - cf auth $CF_STAGING_USER $CF_STAGING_PASSWORD
-      - cf target -o govau -s platform
+      - cf api https://api.system.y.cld.gov.au
+      - cf auth $CF_USER_STAGING $CF_PASSWORD_STAGING
+      - cf target -o dta -s govau
       - curl https://api.rollbar.com/api/1/deploy/ -F access_token=$ROLLBAR_ACCESS_TOKEN
           -F environment=production -F revision=`git log -n 1 --pretty=format:"%H"`
           -F local_username=`whoami`
-      - ./bin/deploy-staging.sh
+      - bundle exec rake assets:precompile
```
m-richo, line 30: @maxious @gordcorp are there any down-sides to running `rake assets:precompile` in CI?  
it's typically done by the CF ruby build pack while staging the application, but devise complains about a missing env-var that doesn't existing in the staging.  Running it hear means CF doesn't try to.

maxious: There are some catches about relying on services but I don't think they should matter to this app http://docs.cloudfoundry.org/buildpacks/ruby/ruby-tips.html#precompile

maxious (approved):

